PR #13, opened 2012-11-28
Groovy devel

egiljones:
I removed all of the code that wasn't necessary for the constraint samplers test from the pr2 kinematics files and congealed the rest down to two cpp/h header pairs - pr2_arm_ik and pr2_arm_kinematics_plugin.  I can go down to one pair by smooshing these together if that makes more sense.


isucan:
I think this is perfectly fine, thanks!

PR #16, opened 2013-01-07
Support for differential IK

marioprats:


sachinchitta:
Looks great.

(1) In the header file, could you please add a comment on which frame the twist is expected to be specified in. 
(2) I think you already have all of these out but there should be no std::couts in the code
(3) Do you have tests for this?

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
```diff
@@ -110,6 +111,12 @@ class JointStateGroup
       within a group. Links that are not in the group are also
       updated, but transforms for joints that are not in the
       group are not recomputed.  */
+  bool setVariableValues(const Eigen::VectorXd& joint_state_values);
```
isucan, line 115: I would rather not have this function. It does a conversion internally anyway. I'd rather put this on the user.

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
```diff
@@ -132,6 +139,9 @@ class JointStateGroup
   
   /** \brief Get current joint values */
   void getVariableValues(std::vector<double>& joint_state_values) const;
+
+  /** \brief Get current joint values */
+  void getVariableValues(Eigen::VectorXd& joint_state_values) const;
```
isucan, line 145: Same here. I would like to remove this function.

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
```diff
@@ -286,6 +296,14 @@ class JointStateGroup
       @param constraint A state validity constraint to be required for IK solutions */  
   bool setFromIK(const std::vector<Eigen::Affine3d> &poses, const std::vector<std::string> &tips, const std::vector<std::vector<double> > &consistency_limits, unsigned int attempts = 0, double timeout = 0.0, const StateValidityCallbackFn &constraint = StateValidityCallbackFn());
 
+  /** \brief Set the joint values from a cartesian velocity applied during a time dt
+   * @param twist a cartesian velocity on the 'tip' frame
+   * @param tip the frame for which the twist is given
+   * @param dt a time interval (seconds)
+   * @param st a secondary task computation function
+   */
+  bool setFromDiffIK(const Eigen::VectorXd &twist, const std::string &tip, const double &dt, const SecondaryTaskCallbackFn &st = SecondaryTaskCallbackFn());
```
isucan: This works for one tip only, correct? The setFromIK() has variants that take multiple tips. Is it difficult to add those? We can do that in a separate pull request. What I would like in this one is to also allow calling setFromDiffIK with the Twist message.

kinematic_state/src/joint_state_group.cpp
```diff
@@ -653,6 +679,64 @@ bool kinematic_state::JointStateGroup::setFromIK(const std::vector<Eigen::Affine
...
+  static const double pinvtoler=1.e-6;
+  double maxsv = 0 ;
+  for (std::size_t i = 0; i < S.rows(); ++i)
+    if (fabs(S(i)) > maxsv) maxsv = fabs(S(i));
+  for (std::size_t i = 0; i < S.rows(); ++i)
+  {
+    //Those singular values smaller than a percentage of the maximum singular value are removed
+    if ( fabs(S(i)) > maxsv * pinvtoler )
+      Sinv(i) = 1.0 / S(i);
+    else Sinv(i) = 0;
```
isucan: use 0.0 and new line after else

kinematic_state/src/joint_state_group.cpp
```diff
@@ -653,6 +679,64 @@ bool kinematic_state::JointStateGroup::setFromIK(const std::vector<Eigen::Affine
...
+
+  //Do the Jacobian moore-penrose pseudo-inverse
+  Eigen::JacobiSVD<Eigen::MatrixXd> svdOfJ(J, Eigen::ComputeThinU | Eigen::ComputeThinV);
+  const Eigen::MatrixXd U = svdOfJ.matrixU();
+  const Eigen::MatrixXd V = svdOfJ.matrixV();
+  const Eigen::VectorXd S = svdOfJ.singularValues();
+
+  Eigen::VectorXd Sinv = S;
+  static const double pinvtoler=1.e-6;
+  double maxsv = 0 ;
```
isucan: use 0.0

kinematic_state/src/joint_state_group.cpp
```diff
@@ -653,6 +679,64 @@ bool kinematic_state::JointStateGroup::setFromIK(const std::vector<Eigen::Affine
...
+  J = eWb * J;
+
+  //Do the Jacobian moore-penrose pseudo-inverse
+  Eigen::JacobiSVD<Eigen::MatrixXd> svdOfJ(J, Eigen::ComputeThinU | Eigen::ComputeThinV);
+  const Eigen::MatrixXd U = svdOfJ.matrixU();
+  const Eigen::MatrixXd V = svdOfJ.matrixV();
+  const Eigen::VectorXd S = svdOfJ.singularValues();
+
+  Eigen::VectorXd Sinv = S;
+  static const double pinvtoler=1.e-6;
```
isucan: use space around =

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
```diff
@@ -50,6 +50,7 @@
 class JointStateGroup;
 
 typedef boost::function<bool(JointStateGroup *joint_state_group, const std::vector<double> &joint_group_variable_values)> StateValidityCallbackFn;
+typedef boost::function<bool(const JointStateGroup &joint_state_group, Eigen::VectorXd &stvector)> SecondaryTaskCallbackFn;
```
isucan: This is not really a callback in the sense that the previous definition is. I would prefer calling this typedef something like SecondaryTaskFn

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
marioprats: What should it do? satisfy one goal pose for each tip?

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
marioprats, line 115: In general I think it is useful to give the user this kind of data in a math library type. One may want to do some computation with it (eg. a controller)

marioprats:
@sachinchitta I will think of a suitable test for this

kinematic_state/include/moveit/kinematic_state/joint_state_group.h
isucan, line 115: Yes, but the conversion is very simple and we enforce a particular math library. If you have strong feelings about keeping this function, I'm fine with it, but I would rather not have it in because it does trivial computation and is used in only one location, as far as i can tell.

PR #22, opened 2013-01-23
Merge gil's distance_field and collision_distance_field updates

kalakris:
What does it take for something like this to get merged?


kalakris:
Shelved for later.

PR #36, opened 2013-01-31
planning_request_adapter: Print useful exception error messages.

kalakris:


isucan:
I already implemented this locally in my branch (will be pushed after friday)

PR #37, opened 2013-02-02
Groovy devel

egiljones:
I'm hoping this is the final deliverable for the distance_field project.  I believe that I've implemented all the cleanup, tests, code refinements, and docs asked for - but definitely look through it and let me know if there still looks like work to do.


acornacorn:
Thanks!  Looks good and seems to work (with FCL and Hybrid).

PR #41, opened 2013-02-06
new computeCartesianPath takes a target pose, interpolates rotation too

marioprats:
Checked with pick up demo and seems to work. Still to check with rotations

### DIFF
--- a/robot_state/src/joint_state_group.cpp
+++ b/robot_state/src/joint_state_group.cpp
@@ -810,7 +810,28 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
   const LinkState *link_state = kinematic_state_->getLinkState(link_name);
   if (!link_state)
     return 0.0;
+
+  //this is the Cartesian pose we start from, and have to move in the direction indicated
+  const Eigen::Affine3d &start_pose = link_state->getGlobalLinkTransform();
   
+  //the direction can be in the local reference frame (in which case we rotate it)
+  const Eigen::Vector3d &rotated_direction = global_reference_frame ? direction : start_pose.rotation() * direction;
+  
+  //The target pose is built by applying a translation to the start pose for the desired direction and distance
+  Eigen::Affine3d target_pose = start_pose;
+  target_pose.translation() += rotated_direction * distance;
+  
+  //call computeCartesianPath for the computed target pose in the global reference frame
+  computeCartesianPath(traj, link_name, target_pose, true, max_step, jump_threshold, validCallback);
+}
+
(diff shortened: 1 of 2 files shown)


marioprats:
Checked rotations too, seems fine to me

robot_state/src/joint_state_group.cpp
```diff
@@ -825,60 +846,65 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
...
-  double last_valid_distance = 0.0;
+
+  double last_valid_percentage = 0.0;
   for (unsigned int i = 1; i <= steps ; ++i)
   {
-    double d = distance * (double)i / (double)steps;
-    Eigen::Affine3d pose = start_pose;
-    pose.translation() = pose.translation() + rotated_direction * d;
+    double percentage = (double)i / (double)steps;
+    Eigen::Quaterniond start_quaternion(start_pose.rotation());
```
isucan: This is the same in every loop iteration, but expensive to compute. Use from outside the loop.

robot_state/src/joint_state_group.cpp
```diff
@@ -825,60 +846,65 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
...
+    Eigen::Quaterniond target_quaternion(rotated_target.rotation());
```
isucan: Move this outside as well.

robot_state/src/joint_state_group.cpp
```diff
@@ -825,60 +846,65 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
...
+    Eigen::Affine3d pose(start_quaternion.slerp((double)i / (double)steps, target_quaternion));
```
isucan: You computed i / steps in percentage. Can you use it here instead?

robot_state/src/joint_state_group.cpp
```diff
@@ -825,60 +846,65 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
...
+    pose.translation() = (i * rotated_target.translation() + (steps - i) * start_pose.translation()) / steps;
```
isucan: It is more efficient to do operations on single doubles rather than on vectors. I think this line is equivalent to:
pose.translation() = percentace \* rotated_target.translation() + (1 - percentage) \* start_pose.translation()

Can you re-check if this works though? It sort of seems like target and pose should be the other way around .... at step 1 i should be closer to the start rather than the target, np?

marioprats:
Fixed. The index is fine, at percentage=0 (and i=0), it's at the start pose.

isucan:
perfect! thanks!

PR #42, opened 2013-02-07
Added more general computeCartesianPath, takes vector of waypoints

marioprats:


robot_state/src/joint_state_group.cpp
```diff
@@ -923,6 +923,30 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
...
+double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotStatePtr> &traj, const std::string &link_name, const std::vector<Eigen::Affine3d> &waypoints,
+                                                          bool global_reference_frame, double max_step, double jump_threshold, const StateValidityCallbackFn &validCallback)
+{
+  double percentage_solved = 0.0;
+  for (std::size_t i = 0; i < waypoints.size(); ++i)
+  {
+    std::vector<RobotStatePtr> waypoint_traj;
+
+    double wp_percentage_solved = computeCartesianPath(waypoint_traj, link_name, waypoints[i], global_reference_frame, max_step, jump_threshold, validCallback);
+    if (wp_percentage_solved == 1)
```
isucan: when comparing doubles, is it better do do things like 
fabs(wp_percentage_solved - 1.0) < std::numeric_limits<double>::epsilon()

robot_state/src/joint_state_group.cpp
```diff
@@ -923,6 +923,30 @@ double robot_state::JointStateGroup::computeCartesianPath(std::vector<RobotState
...
+    {
+      percentage_solved = ((double)i + 1) / (double)waypoints.size();
```
isucan: do (double)(i+1) instead of (double)i + 1

PR #51, opened 2013-03-21
state lookup utility for robot_trajectory

aleeper:
An implementation of "get the robot state at some duration from the start of the trajectory".
This is what I use with my teleop stuff, so... it seems to work :)

### DIFF
--- a/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -104,8 +104,9 @@ public:
     return duration_from_previous_;
   }
 
-  /** @brief Returns the duration after start that a waypoint will be reached.
-   *  @param The waypoint index.
+  /** @brief  Returns the duration after start that a waypoint will be reached.
+   *  @param  The waypoint index.
+   *  @return The duration from start; retuns -1.0 if index is out of range.
    */
   double getWaypointDurationFromStart(std::size_t index) const;
   
@@ -186,21 +187,19 @@ public:
   /** @brief Finds the waypoint indicies before and after a duration from start.
    *  @param The duration from start.
    *  @param The waypoint index before the supplied duration.
-   *  @param The waypoint index after the supplied duration.
-   *  @param The progress (0 to 1) between the two waypoints, based on time.
+   *  @param The waypoint index after (or equal to) the supplied duration.
+   *  @param The progress (0 to 1) between the two waypoints, based on time (not based on joint distances).
    */
-  void findWayPointIndicesForDurationAfterStart(const double& duration, int& before, int& after, double &blend);
+  void findWayPointIndicesForDurationAfterStart(const double& duration, int& before, int& after, double &blend) const;
 
-  /** @brief Gets a robot state corresponding to a supplied duration from start for the trajectory.
+  // TODO support visitor function for interpolation, or at least different types.
+  /** @brief Gets a robot state corresponding to a supplied duration from start for the trajectory, using linear time interpolation.
    *  @param The duration from start.
-   *  @param A flag: if true, attempts linear interpolation; else uses first waypoint after input duration.
    *  @param The resulting robot state.
    *  @return True if state is valid, false otherwise (trajectory is empty).
    */
   bool getStateAtDurationFromStart(const double request_duration,
-                                   const bool interpolate_between_waypoints,
-                                   robot_state::RobotStatePtr& output_state,
-                                   double& actual_duration);
+                                   robot_state::RobotStatePtr& output_state) const;
 
 private:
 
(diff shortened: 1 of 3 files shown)


robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -177,7 +182,26 @@ class RobotTrajectory
 
   void unwind();
   void unwind(const robot_state::RobotState &state);
-  
+
+  /** @brief Finds the waypoint indicies before and after a duration from start.
+   *  @param The duration from start.
+   *  @param The waypoint index before the supplied duration.
+   *  @param The waypoint index after the supplied duration.
+   *  @param The progress (0 to 1) between the two waypoints, based on time.
```
isucan: what if the duration points exactly to the waypoint K? will i have K == before or K == after ?

robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -177,7 +182,26 @@ class RobotTrajectory
...
+  /** @brief Gets a robot state corresponding to a supplied duration from start for the trajectory.
+   *  @param The duration from start.
+   *  @param A flag: if true, attempts linear interpolation; else uses first waypoint after input duration.
+   *  @param The resulting robot state.
+   *  @return True if state is valid, false otherwise (trajectory is empty).
+   */
+  bool getStateAtDurationFromStart(const double request_duration,
+                                   const bool interpolate_between_waypoints,
+                                   robot_state::RobotStatePtr& output_state,
+                                   double& actual_duration);
```
isucan: can this function be const?

robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -177,7 +182,26 @@ class RobotTrajectory
...
+  void findWayPointIndicesForDurationAfterStart(const double& duration, int& before, int& after, double &blend);
+
+  /** @brief Gets a robot state corresponding to a supplied duration from start for the trajectory.
+   *  @param The duration from start.
+   *  @param A flag: if true, attempts linear interpolation; else uses first waypoint after input duration.
+   *  @param The resulting robot state.
+   *  @return True if state is valid, false otherwise (trajectory is empty).
+   */
+  bool getStateAtDurationFromStart(const double request_duration,
+                                   const bool interpolate_between_waypoints,
```
isucan: would we ever not want to interpolate? It seems that if not, we can just get the waypoint. Maybe we can remove the flag and always do interpolation if needed?

robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -177,7 +182,26 @@ class RobotTrajectory
...
   void unwind(const robot_state::RobotState &state);
-  
+
+  /** @brief Finds the waypoint indicies before and after a duration from start.
+   *  @param The duration from start.
+   *  @param The waypoint index before the supplied duration.
+   *  @param The waypoint index after the supplied duration.
+   *  @param The progress (0 to 1) between the two waypoints, based on time.
+   */
+  void findWayPointIndicesForDurationAfterStart(const double& duration, int& before, int& after, double &blend);
```
isucan: can this function be const?

robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -306,3 +306,83 @@ void robot_trajectory::RobotTrajectory::setRobotTrajectoryMsg(const robot_state:
...
+  if(after == before)
+    blend = 1.0;
+  else
+    blend = (duration - before_time) / duration_from_previous_[index];
+}
+
+double robot_trajectory::RobotTrajectory::getWaypointDurationFromStart(std::size_t index) const
+{
+  if(index >= duration_from_previous_.size())
+    return -1.0; // or something else? should we throw an exception?
```
isucan, line 344: -1.0 is fine. Just add a comment in the docs

robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -306,3 +306,83 @@ void robot_trajectory::RobotTrajectory::setRobotTrajectoryMsg(const robot_state:
...
+
+  double time = 0;
+  for(size_t i = 0; i <= index; i++)
```
isucan: can we use ++i and std::size_t (full ns prefix)

robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -177,7 +182,26 @@ class RobotTrajectory
...
+
+  /** @brief Gets a robot state corresponding to a supplied duration from start for the trajectory.
+   *  @param The duration from start.
+   *  @param A flag: if true, attempts linear interpolation; else uses first waypoint after input duration.
+   *  @param The resulting robot state.
+   *  @return True if state is valid, false otherwise (trajectory is empty).
+   */
+  bool getStateAtDurationFromStart(const double request_duration,
+                                   const bool interpolate_between_waypoints,
+                                   robot_state::RobotStatePtr& output_state,
```
isucan: I think this argument type should be  const robot_state::RobotStatePtr&

PR #52, opened 2013-03-23
Added method to get joint type as string

davetcoleman:


davetcoleman:
Ah I reformatted the code again, sorry. Its on line 97

Issue #35, opened 2013-01-30
make iterative time parameterization tool use KinematicTrajectory


isucan:

isucan:
In the moveit_core package, the trajectory_processing library includes the IterativeParabolicTimeParameterization class; That class computes time staps for the waypoints of the robot along a given trajectory. 

The issue that needs to be resolved is that the computeTimeStamps() function should operate directly on the robot_trajectory::RobotTrajectory class; currently, it instead converts the trajectory to a message (trajectory_msgs::JointTrajectory) then calls the computeTimeStamps() variant  that operates on the message and then converts back to a RobotTrajectory.

What needs to be done is that the computation needs to be performed directly on the RobotTrajectory class, in place (no copying of states) and the version of computeTimeStamps() that operates on the message should be implemented in terms of the function that operates directly on RobotTrajectory, using conversion functions.

tulku:
We have started working on this and committed the first changes to my fork.

I have a question, in the bug report you state that the original JointTrajectory should still be available as wrappers of the RobotTrajectory ones, whoever, they are not being use by anyone right one. Are they really necessary? (I've made them private and everything compiles fine).

aleeper:
I think we should keep them; it makes it easier for folks who might already have a JointTrajectory and want to manually call the time parameterization in their own code (which I have done sometimes).

isucan:
sorry for the late reply - pretty bad internet at the conf and the schedule was a bit hectic. 
the code looks great, thanks! the functions you mention are indeed not used anywhere, but i agree with adam about keeping them. the implementation should be simple anyway,  so there is no problem having them. we will just include in the docs the fact that conversions are performed and the use is not encouraged.

tulku:
Ok, I'll leave them public using the conversions. Just asked to avoid possible dead code. Thanks!

tulku:
Hi!

I have been spending quite some time now finishing this patch and run into a problem.

Part of the process that's done to the trajectory_msgs::JointTrajectory is calling a function called updateTrajectory which is suppose to "Takes the time differences, and updates the values in the trajectory". However this function also makes some velocity (although the code is commented) and acceleration calculations. This new values are stored in the JointTrajectory message.

[...]

PR #62, opened 2013-05-16
issue #35 almost complete

tulku:
All the timestamp calculations are now done in place in the robot_trajectory::RobotTrajectory object.

To close the issue the new wrappers that accept a JointTrajectory message should be written.

### DIFF
--- a/trajectory_processing/src/iterative_time_parameterization.cpp
+++ b/trajectory_processing/src/iterative_time_parameterization.cpp
@@ -100,47 +100,64 @@ void IterativeParabolicTimeParameterization::printStats(const trajectory_msgs::J
 
 // Applies velocity
 void IterativeParabolicTimeParameterization::applyVelocityConstraints(trajectory_msgs::JointTrajectory& trajectory,
+                                                                      robot_trajectory::RobotTrajectory& rob_trajectory,
                                                                       const std::vector<moveit_msgs::JointLimits>& limits,
                                                                       std::vector<double> &time_diff) const
 {
-
   // aleeper: This function computes the minimum time needed for each trajectory
   //          point by (forward_difference/ v_max), and recording the minimum time in
   //          the time_diff vector, which will be used later.
-  const unsigned int num_points = trajectory.points.size();
-  const unsigned int num_joints = trajectory.joint_names.size();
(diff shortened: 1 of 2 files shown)


trajectory_processing/src/iterative_time_parameterization.cpp
```diff
@@ -439,72 +508,75 @@ void IterativeParabolicTimeParameterization::applyAccelerationConstraints(const
...
   {
-    //logInform("We seem to have velocity data; populating...");
-    for (std::size_t i = 0; i < start_state.joint_state.name.size(); ++i)
+    for (std::size_t j = 0; j < active_joints.size(); ++j)
     {
-      //logInform("joint: [%s], velocity: [%.3f]", start_state.joint_state.name[i].c_str(), start_state.joint_state.velocity[i] );
-      velocity_map[start_state.joint_state.name[i]] = start_state.joint_state.velocity[i];
+      jst = trajectory.getWayPointPtr(i)->getJointState(active_joints[j]);
+      if (!jst->getVelocities().empty())
+        velocity_map[active_joints[j]] = jst->getVelocities()[0];
```
tulku, line 558: This should be checked. Probably we should keep the velocities of the first waypoint.

trajectory_processing/src/iterative_time_parameterization.cpp
```diff
@@ -439,72 +508,75 @@ void IterativeParabolicTimeParameterization::applyAccelerationConstraints(const
...
+
+  const robot_model::JointModelGroup *group = trajectory.getGroup();
+  const std::vector<const robot_model::JointModel*> &jnt = group->getJointModels();
+  const std::vector<moveit_msgs::JointLimits> &limits = trajectory.getGroup()->getVariableLimits();
+
+  // \todo this lib does not actually work properly when angles wrap around, so we need to unwind the path first
+  trajectory.unwind();
+
+  for (int i=0; i < jnt.size(); i++)
+    active_joints.push_back(jnt[i]->getName());
```
tulku, line 543: Using the active_joints vector and then getting the joint positions using the variable map of the robot state is most certainly not the most efficient way of getting this information.

PR #63, opened 2013-05-16
Add bounding spheres to robot_model::RobotModel

acornacorn:
DEPENDS ON: https://github.com/ros-planning/srdfdom/pull/1
DEPENDS ON: https://github.com/ros-planning/geometric_shapes/pull/3

### DIFF
--- a/robot_model/src/robot_model.cpp
+++ b/robot_model/src/robot_model.cpp
@@ -671,7 +671,7 @@ robot_model::JointModel* robot_model::RobotModel::buildRecursive(LinkModel *pare
   if (joint->getType() == JointModel::REVOLUTE && static_cast<const RevoluteJointModel*>(joint)->isContinuous())
     continuous_joint_model_vector_const_.push_back(joint);
   joint->parent_link_model_ = parent;
-  joint->child_link_model_ = constructLinkModel(link, parent_map);
+  joint->child_link_model_ = constructLinkModel(link, parent_map, srdf_model);
   link_model_map_[joint->child_link_model_->name_] = joint->child_link_model_;
   joint->child_link_model_->tree_index_ = link_model_vector_.size();
   link_model_vector_.push_back(joint->child_link_model_);
@@ -842,7 +842,7 @@ static inline Eigen::Affine3d urdfPose2Affine3d(const urdf::Pose &pose)
 
 }
 
-robot_model::LinkModel* robot_model::RobotModel::constructLinkModel(const urdf::Link *urdf_link, const std::map<const urdf::Link*, std::pair<const urdf::Link*, const urdf::Joint*> > &parent_map)
+robot_model::LinkModel* robot_model::RobotModel::constructLinkModel(const urdf::Link *urdf_link, const std::map<const urdf::Link*, std::pair<const urdf::Link*, const urdf::Joint*> > &parent_map, const srdf::Model &srdf_model)
 {
   LinkModel *result = new LinkModel();
   result->name_ = urdf_link->name;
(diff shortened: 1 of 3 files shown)


robot_model/src/robot_model.cpp
```diff
@@ -877,6 +877,35 @@ static inline Eigen::Affine3d urdfPose2Affine3d(const urdf::Pose &pose)
...
+      result->collision_sphere_radii_.push_back(sphere_it->radius_);
+    }
+  }
+  if (result->collision_sphere_radii_.empty() && result->shape_)
+  {
+    // by default use a single sphere that bounds the entire link
+    result->collision_sphere_centers_.resize(1);
+    result->collision_sphere_radii_.resize(1);
+    shapes::computeShapeBoundingSphere(result->shape_.get(), result->collision_sphere_centers_[0], result->collision_sphere_radii_[0]);
+    if (result->collision_sphere_radii_[0] == 0.0)
```
isucan, line 896: This looks weird to me; I guess it relates to the 0 sized spheres in the srdfdom

acornacorn, line 896: The idea is to remove any 0-sized spheres. In the robot model any link without collision spheres will be ignored for distance field collision.

robot_model/src/robot_model.cpp
```diff
@@ -877,6 +877,35 @@ static inline Eigen::Affine3d urdfPose2Affine3d(const urdf::Pose &pose)
...
+    {
+      result->collision_sphere_centers_.clear();
+      result->collision_sphere_radii_.clear();
+    }
+  }
+  if (result->collision_sphere_radii_.size() == 1 && result->collision_sphere_radii_[0] == 0.0)
+  {
+    // if only one sphere and radius is 0 then skip sphere collision for this link.
```
isucan, line 904: This does not make sense to me; I think we should leave spheres be as they are. If the size is 0, the it will be a bit less efficient. We should not worry about this.

acornacorn, line 904: I disagree.  Collision checking is all about efficiency.  When would it ever make sense to have a 0-radius sphere?

acornacorn, line 904: Actually, if you really don't like changing it here then I can eliminate unused spheres later when constructing the CollisionRobotDistanceField object.  Would that be better?

The question then is what does a radius==0 sphere mean?  With this code it is clear that such a sphere is ignored (because we eliminate it).  But if it does not get eliminated then it could have 2 possible meanings:

A) a very small object
B) a nonexistant object

I really do not want it to mean A.  But if we allow them to exist then they can be confused to mean B in future code.

isucan, line 904: github is funny. I could swear it posted an answer to the before.
I am fine with removing the spheres, and i think it should be done here. I was just under the impression you had 0 sized spheres be of some special type. You should check for radius < std::numeric_limits<double>() instead of 0.

Also, since size() == 1 always in this initalization, the if statement can be simplified

acornacorn, line 904: OK.  I'll change it to check std::numeric_limits() instead of 0.  

Note that size is not always 1.  It is initialized to 1 if no spheres exist, but if the SRDF has >1 spheres then the list here will too.  I am counting on the srdf code to ensure that either there are no 0-radius spheres or else only a single one.  If the srdf has a 0-radius sphere that is a bug.  But I want to be robust.  The easiest way to be robust is eliminate the sphere if there is a single 0-radius sphere, but just leave it if there are >1 spheres and some have 0 radius.  It would be more robust to go through the list of spheres and eliminate any that are 0-radius, but that would make dealing with IDs difficult (I'll send an email to explain the ids and ask your opinion about them...).  

Actually maybe best to wait for me to send the email and try to explain there...

isucan, line 904: i think the srdf parser should remove all 0-sized spheres; there will just be no spheres for a particular link in that case.

acornacorn:
Decided to leave spheres in srdf and not add to RobotModel.  Closing pull request.

PR #65, opened 2013-05-23
remove direction from CollisionResult

acornacorn:
Direction was never being filled in and never being checked.  It is unclear
what this would represent.  Deleting it.

### DIFF
--- a/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -135,7 +135,6 @@ namespace collision_detection
   {
     CollisionResult() : collision(false),
                         distance(std::numeric_limits<double>::max()),
-                        direction(0.0, 0.0, 0.0),
                         contact_count(0)
     {
     }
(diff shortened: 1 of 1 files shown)


isucan:
The intention was to specify the vector along which the minimum distance would be achieved. But we need a reference point too, so this is incomplete. Since we do not use it now, I agree we should just remove it.

aleeper:
If it's actually being removed, then we should also remove it from CollisionRequest.

BUT, I would ask if we actually want to do this. It seems like it would be more correct to keep the distance member and add the point, as some collision detectors might implement this. (e.g. how does collision distance field work?). Even though distance might not be that useful for randomized planning, predicting impending contact with a distance value could really help some of the control-based methods in teleoperation.

isucan:
I agree this is useful information. It is also pretty easy to add back I think. I just though we wait for a particular application to add this back. 
Acorn, if it is almost trivial for you, we can add it as two points (the ones that give the distance) instead of the vector between them.
There is no member in the CollisionRequest that relates to the distance vector that just got removed.

acornacorn:
It is not hard to add, but lets see how it is useful to clients before we
add it back.  No sense adding some obscure feature that no one will use.
 It you know how you would use this let me know the details.

PR #76, opened 2013-06-26
Added ability to change planner configurations in the interface

davetcoleman:
This is for parameter sweeping


planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -45,6 +45,15 @@
 namespace planning_interface
 {
 
+/// Holds settings specific for each planner, as found in i.e. ompl_planning.yaml
```
isucan: This file cannot be specific to ompl; We should remove the reference above

planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -45,6 +45,15 @@
...
+struct PlanningConfigurationSettings
+{
+  std::string                        name;
```
isucan, line 55: We should document things here; say how the settings are used. 
I suggest the following policy: settings with unknown keys are ignored.
Settings for unknown groups are ignored.

planning_interface/include/moveit/planning_interface/planning_interface.h
```diff
@@ -78,6 +87,12 @@ class Planner
   
   /// Determine whether this plugin instance is able to represent this planning request
   virtual bool canServiceRequest(const MotionPlanRequest &req)  const = 0;
+
+  /// Settings for planners to use for their specific algorithms
+  virtual void setPlanningConfigurations(const PlanningConfigurationMap &pcs) const = 0;
+
+  /// Get settings for planners to use for their specific algorithms, so that they can then be modified
+  virtual const PlanningConfigurationMap getPlanningConfigurations() const = 0;
```
isucan, line 101: should we return by value here? Or should we fill an arg? Maybe this is fine as it is.

planning_interface/include/moveit/planning_interface/planning_interface.h
davetcoleman, line 55: I don't completely remember how this struct works - it came from OMPL and I just added it to MoveIt. I have documented what I can.

planning_interface/include/moveit/planning_interface/planning_interface.h
davetcoleman, line 101: no opinion.

PR #81, opened 2013-07-24
Adds ability to find attached bodies for a group

sachinchitta:


planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -230,6 +230,13 @@ class PlanningScene : private boost::noncopyable,
   /** \brief Get the set of fixed transforms from known frames to the planning frame */
   robot_state::Transforms& getTransformsNonConst();
 
+  /** \brief Get the shared ptr to the transforms */
+  const robot_state::TransformsPtr& getTransformsPtr() const
```
isucan, line 234: This function is not consistent with the rest of the API in PlanningScene. Can you point me to where it is used? Maybe we can do things differently.

robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -210,6 +210,9 @@ class RobotState
   /** \brief Get all bodies attached to the model corresponding to this state */
   void getAttachedBodies(std::vector<const AttachedBody*> &attached_bodies) const;
 
+  /** \brief Get all bodies attached to a group corresponding to this state */
+  void getAttachedBodies(const std::string &group_name, std::vector<const AttachedBody*> &attached_bodies) const;
```
isucan: This should move to be part of a JointStateGroup

planning_scene/include/moveit/planning_scene/planning_scene.h
sachinchitta, line 234: This is used in motion_planning_frame.cpp to get a transforms ptr to pass on to semantic_world.

isucan, line 234: I think we should pass a PlanningScenePtr to the semantic world instead of the Transforms object. Then we can remove this version of the function.

robot_state/src/joint_state_group.cpp
```diff
@@ -1143,3 +1143,31 @@ bool robot_state::JointStateGroup::getJacobian(const std::string &link_name,
   }
   return std::pair<double,int>(distance,index);
 }
+
+void robot_state::JointStateGroup::getAttachedBodies(std::vector<const robot_state::AttachedBody*> &attached_bodies) const
+{
+  logDebug("Getting attached bodies for group %s", getJointModelGroup()->getName().c_str());
```
isucan, line 1149: I think this is basically spam :) I will remove it after the merge.

robot_state/src/joint_state_group.cpp
```diff
@@ -1143,3 +1143,31 @@ bool robot_state::JointStateGroup::getJacobian(const std::string &link_name,
...
+  
+  std::vector<const robot_state::AttachedBody*> robot_attached_bodies;
+  getRobotState()->getAttachedBodies(robot_attached_bodies);
+  
+  std::vector<const robot_state::AttachedBody*>::const_iterator it;  
+  for(it = robot_attached_bodies.begin(); it != robot_attached_bodies.end(); ++it)
```
isucan, line 1155: I don't like this implementation. We should just loop over the links of this group instead.

PR #82, opened 2013-07-25
enable RTTI for CollisionRequest

acornacorn:
This allows passing a subclass of CollisionRequest to request more specific results from a specific collision checker.


isucan:
do we need this in groovy-devel? If not, I would prefer adding this to hydro-devel instead

isucan:
actually this is small enough; i'll cherry-pick the commit on hydro-devel as well

PR #83, opened 2013-08-02
Assert replaced by return in IKTconstraints check.

petrikvladimir:
This make moveit more stable for inaccurate IK solvers. Since IKT is computed for randomly chosen points, inaccurate IKT can return dissatisfied result for points near the border of constraints region.


isucan:
This also makes accurate solvers spend significantly more time per IK request. That is basically why we do assert() there -- so that the code does not get generated in release mode. If the solver is inaccurate, it could perhaps add a check itself and return false. Or, we can add a boolean flag in the kinematics base class where a solver can declare it will not always satisfy requests correctly. That is a bit weird though ...

petrikvladimir:
Thanks for the explanation, I did not known the assert is not generated in the relase mode. Inaccuracy of our IK solver is small so it will not be problem in real application.

PR #86, opened 2013-09-09
Added new function to get a robot model by its index

davetcoleman:
There is probably another function that has this same functionality, but I couldn't find it/didn't like its name.

I also created a new version of getVariableIndex that returns false if it can't find it, instead of throwing an error.

This is necessary for https://github.com/ros-planning/moveit_ros/pull/308


isucan:
Thanks! I included some of the functions you added.

PR #103, opened 2013-11-05
fixing broken tests for changes in robot_state

sachinchitta:
This fixes the test - i.e. they now compile. Lots of tests are still failing but it would be useful for this PR to be merged now so other people can start looking at the tests themselves.


isucan:
Thank you!

PR #105, opened 2013-11-08
Compute associated transforms bug fixed.

petrikvladimir:


isucan:
I did not have time to test. Have you verified this ?

petrikvladimir:
Yes, I verified it. Here is the code and the result with and without the fix:
https://bitbucket.org/voop/clopema/src/c6bb24f64b7b4c3f311d90ec87bb7b3327336acf/clopema_pub_sandbox/src/robot_state_bug.cpp?at=master

PR #106, opened 2013-11-08
RobotState: Made transforms and attached bodies mutable to ease const-ness.

hersh:
PLEASE DO NOT CASUALLY MERGE THIS.

This is posted mostly for discussion as an easy way to view a proposed change.

The const versions of getGlobalLinkTransforms() previously didn't do the same thing as the non-const versions... the const versions did not update the transforms, thus giving wrong answers (in release build) or failing an assertion in debug build.

This pull request shows a possible way around this.

What do people think?

### DIFF
--- a/robot_state/include/moveit/robot_state/robot_state.h
+++ b/robot_state/include/moveit/robot_state/robot_state.h
@@ -898,7 +898,7 @@ as the new values that correspond to the group */
   void updateCollisionBodyTransforms();
   
   /** \brief Update the reference frame transforms for links. This call is needed before using the transforms of links for coordinate transforms. */
-  void updateLinkTransforms();
+  void updateLinkTransforms() const;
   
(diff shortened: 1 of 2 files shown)


isucan:
I do not like this change -- this makes it possible to have two threads that write to the same area of memory with no locking. Adding locks will make things too slow. The only thing we can do (and I think should do) is to make things crash when code is compiled with optimizations as well (aka, not use assert(), but a check that is not removed at a compile time instead).

hersh:
How does this relate to thread access?

hersh:
Ah, I guess because someone might see the "const" on a function and presume that it is therefore thread-safe.

isucan:
exactly :)

hersh:
Closing this pull request, it should not be merged.  We'll clarify the API some other way.

PR #127, opened 2013-12-03
fix for computing jacobian when the root_joint is not an active joint

sachinchitta:


robot_state/src/robot_state.cpp
```diff
@@ -975,7 +975,7 @@ bool moveit::core::RobotState::getJacobian(const JointModelGroup *group, const L
     return false;
   }
 
-  const robot_model::JointModel* root_joint_model = group->getJointRoots()[0];
+  const robot_model::JointModel* root_joint_model = group->getJointModels()[0];//group->getJointRoots()[0];
```
isucan, line 978: I am assuming you are doing this to account for the case when a root joint is fixed, correct?

hersh, line 978: yes, he is.

hersh, line 978: (he is here now)

sachinchitta:
This should also work for the case when the root joint is a mimic joint.

robot_state/src/robot_state.cpp
isucan, line 978: sounds good. Just checking that was indeed the intention :)

PR #130, opened 2014-01-06
Add kinematic options when planning for CartesianPath

skohlbr:
Probably an earlier oversight. kinematic options forwarded to computeCartesianPath would not be forwarded. This lead to torso movement of our Atlas when we didn´t want it, but was fixed using the proposed change.


skohlbr:
As proposed in mail exchange by Sachin.

isucan:
Thanks, this looks good!

PR #133, opened 2014-01-09
Add comments to RobotState

acornacorn:


robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -880,11 +880,32 @@ class RobotState
       setVariableVelocities(msg.name, msg.velocity);
   }
   
+  /** \brief Set all joints to their default positions */
```
isucan, line 883: We should also say that this is 0 if 0 is within bounds, or middle of the range, if 0 is not in the range.

acornacorn, line 883: Thanks!  Added.

PR #138, opened 2014-01-19
fix mix-up comments, use getCollisionRobotUnpadded() since this function...

k-okada:
fix comments, in order to avoid confusion, the function checkCollisionUnpadded intentionally uses getCollisionRobotUnpadded()


sachinchitta:
Thanks Kei!

PR #144, opened 2014-02-13
Fixed asymmetry between planning scene read and write.

hersh:
Previously when reading a planning scene, every object would have a " " space prepended onto the name.  If you loaded and saved and loaded and saved the same scene, you would get more and more spaces on the name.  The way it's presented in the GUI does not show up the spaces clearly, so it's not casually evident until you start trying to automate things and finding that the object you want is not there... instead it's called " object".


isucan:
+1

PR #145, opened 2014-02-13
KinematicsMetrics crashes when called with non-chain groups.

hersh:
This change makes it so the KinematicsMetrics class returns false (failure) instead of causing exceptions to be thrown which no one catches and thus crashes your program.

The ultimate cause is that RobotState::getJacobian() does not have a proper error return value, it throws an exception instead.  That API should ultimately be changed, but this change avoids the crash with no API change and is thus less of a big deal in my opinion.

### DIFF
--- a/kinematics_metrics/src/kinematics_metrics.cpp
+++ b/kinematics_metrics/src/kinematics_metrics.cpp
@@ -103,6 +103,12 @@ bool KinematicsMetrics::getManipulabilityIndex(const robot_state::RobotState &st
                                                double &manipulability_index,
                                                bool translation) const
 {
+  // state.getJacobian() only works for chain groups.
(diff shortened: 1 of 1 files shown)


isucan:
+1

davetcoleman:
I just ran into this problem! Tested. This fix helps my goal of non-chain IK solving, thanks!

sachinchitta:
@hersh can you confirm that travis went through and I will merge this.

hersh:
The travis problem is caused by this: https://github.com/ros/rosdistro/pull/3072 .  It will be breaking all our travis builds until they get their fix pushed out.

I personally vouch for this change, I've been using it for a while now, so I think is safe to merge.

PR #146, opened 2014-02-14
Debug and documentation

davetcoleman:
No functional changes in this PR:
- There are a lot of debug messages that print continuously or frequently. I have added prefixes for these per this new feature: https://github.com/ros/rosconsole_bridge/pull/4
- I've started documentation on constraint samplers, though its by no means done
- I corrected a few doxygen comments


isucan:
+1

PR #147, opened 2014-02-14
Added new virtual getName() function to constraint samplers

davetcoleman:
This is for easier debugging and plugin management. I don't know of anyone who has created their own constraint samplers yet, but I'm in the process of implementing one. Its hard to debug MoveIt! when plugins have no way to identify themselves into terminal, so I've added this function.

Because it modified the constraint_sampler plugin header by adding a new virtual function, it might be better to merge this into a new indigo-branch for ABI compatibility. I'm not an expert on this.

Also, there are some documentation fixes.

### DIFF
--- a/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
+++ b/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
@@ -167,8 +167,7 @@ public:
 
   /**
    * \brief Samples given the constraints, populating \e state.
-   * The value DEFAULT_MAX_SAMPLING_ATTEMPTS will be passed in
-   * as the maximum number of attempts to make to take a sample.
+   * This function allows the parameter max_attempts to be set.
    *
    * @param state The state into which the values will be placed. Only values for the group are written. The same state is used as reference if needed.
    * @param [in] max_attempts The maximum number of times to attempt to draw a sample.  If no sample has been drawn in this number of attempts, false will be returned.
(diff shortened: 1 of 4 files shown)


constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
```diff
@@ -257,6 +256,13 @@ class ConstraintSampler
     verbose_ = flag;
   }
 
+  /**
+   * \brief Get the name of the constraint sampler, for debugging purposes
+   * should be in CamelCase format.
+   * \return string of name
+   */
+  virtual std::string getName() = 0;
```
isucan: This imposes a constraint on the implementation, but it would be nice if the type was const std::string& and the member function itself is const. You can use a static var in the functions below.

isucan:
+1

constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
davetcoleman: I believe I made your requested change. Should the static const variable be in or out of the function?

hersh:
I'm disappointed that I didn't comment on this at the time... I hate this thing of returning const references to strings.  It is a performance improvement for a debug-write.  It is a hassle for implementers, as you see in the code, there are 10 new lines of C++ (including comments) where there should be (maybe) 1 line:

```
std::string getName() const { return "UnionConstraintSampler"; }
```

I say "maybe" because if you need to find out at run-time what the type of something is for debugging, just use typeid() in your client code:

```
#include <iostream>
#include <typeinfo>

class Foo
{
  int bar;
};

int main(int argc, char** argv)
{
  Foo* foo = new Foo;
  std::cout << "type of foo is '" << typeid(foo).name() << "'" << std::endl;
}
```

This outputs 

```
type of foo is 'P3Foo'
```

davetcoleman:
> It is a performance improvement for a debug-write

Good point.

> 10 new lines of C++ (including comments)

Shouldn't you always have proper doxygen comments though?

> use typeid() in your client code

I actually didn't know this trick, thanks!

PR #148, opened 2014-02-18
Improved error output for state conversion

davetcoleman:
Fixed silent error when converting a `moveit_msgs::RobotState` message to a `RobotState` object. Other failures in robot_state/conversions have failure messages, so this should too.

### DIFF
--- a/robot_state/src/conversions.cpp
+++ b/robot_state/src/conversions.cpp
@@ -103,6 +103,7 @@ static bool _multiDOFJointsToRobotState(const sensor_msgs::MultiDOFJointState &m
       }
     else
       error = true;
+
     if (error)
       logWarn("The transform for multi-dof joints was specified in frame '%s' but it was not possible to transform that to frame '%s'",
               mjs.header.frame_id.c_str(), state.getRobotModel()->getModelFrame().c_str());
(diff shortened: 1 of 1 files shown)


sachinchitta:
Aah, this PR is a regression - did not realize that earlier - it will have to be pulled back. Looks like the meaning of "missing" values is being misinterpreted here since the incoming const state message can always have missing values (which is what is being checked for). 

@isucan - could you check and please confirm that I am correct.

isucan:
You are correct Sachin. The error message is incorrect and we need to pull this one back.

davetcoleman:
I apologize for the regression, though I still don't fully understand... the header for `robotStateMsgToRobotState` says it returns false if it "failed for any reason", so the error message I added informs the user why it failed. 

I believe you are saying that it is ok to fail and we should not have an error message? I think the header doxygen should then be reworded to say something like "returns false if not all frames of the robot state are provided" and then take out my error message. That is to say, define the return state as not being a failure.

sachinchitta:
Its fine - I think there was a bug the way it was setup originally anyways - I pushed a new PR that should fix this #152.

davetcoleman:
Ah, so now it doesn't care about missing links, that was a bug indeed.

PR #149, opened 2014-02-20
KinematicsBase support for multiple tip frames and IK requests with multiple poses

davetcoleman:
This PR only modifies KinematicsBase functionality to make review easier, though changes will be needed elsewhere in robot_state and moveit_ros to enable full non-chain IK support.

This introduces 3 new functions that optionally allow multiple tips and poses to be passed into KinematicsBase:

`searchPositionIK()` - allows a vector of poses to be passed in
`setValues()` - allows a vector of frame tips to be passed in
`initialize()` - allows a vector of frame tips to be passed in
`getTipFrames()` - returns a vector of tips instead of just one

The old `std::string tip_frame_` member variable is a bit messy in that we cannot remove it without breaking older code like generated IKFast solver code. Thus, this variable must remain in for the time being. The results in a little bit of duplicate information if you are using the IK solver in the normal case of a chain solver in that your tip frame name is saved in both the string and in the vector. I've done my best to document the deprecated status of this variable throughout the code and in the future I think it should be removed. 

I have tested these modifications against KDL, IKFast, and our custom full body IK solver.

The following PRs depends on these changes:
[Allow planning groups to have more than one tip](https://github.com/ros-planning/moveit_ros/pull/431)
[Allow joint model group to have use IK solvers with multiple tip frames](https://github.com/ros-planning/moveit_core/pull/155)

### DIFF
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -32,7 +32,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
-/* Author: Sachin Chitta */
+/* Author: Sachin Chitta, Dave Coleman */
 
 #ifndef MOVEIT_KINEMATICS_BASE_KINEMATICS_BASE_
 #define MOVEIT_KINEMATICS_BASE_KINEMATICS_BASE_
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -404,13 +404,23 @@ public:
 
 
   /**
-   * \brief Only kinematic solvers know if they support a particular kinematic configuration, put the burden on them
-   *        to return an explanation of why they are unable to service a joint model group
-   *        Default implementation for legacy solvers provided
-   * \param JointModelGroup - the planning group being proposed to be solved by this IK solver
-   * \return a string specifying the reason the joint model group is not supported, or an empty string if supported
+   * \brief Check if this solver supports a given JointModelGroup.
+   *
+   * Override this function to check if your kinematics solver
+   * implementation supports the given group.
+   *
+   * The default implementation just returns jmg->isChain(), since
+   * solvers written before this function was added all supported only
+   * chain groups.
+   *
+   * \param jmg the planning group being proposed to be solved by this IK solver
+   * \param error_text_out If this pointer is non-null and the group is
+   *          not supported, this is filled with a description of why it's not
+   *          supported.
+   * \return True if the group is supported, false if not.
    */
-  virtual const std::string supportsGroup(const moveit::core::JointModelGroup *jmg) const;
+  virtual const bool supportsGroup(const moveit::core::JointModelGroup *jmg,
+                                   std::string* error_text_out = NULL) const;
 
   /**
    * @brief  Set the search discretization
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -43,6 +43,14 @@
 #include <console_bridge/console.h>
 #include <string>
 
+namespace moveit
+{
+namespace core
+{
+class JointModelGroup;
+}
+}
+
 /** @brief API for forward and inverse kinematics */
 namespace kinematics
 {
@@ -393,6 +401,16 @@ public:
    * @brief  Return all the link names in the order they are represented internally
    */
   virtual const std::vector<std::string>& getLinkNames() const = 0;
+
+
+  /**
+   * \brief Only kinematic solvers know if they support a particular kinematic configuration, put the burden on them
+   *        to return an explanation of why they are unable to service a joint model group
+   *        Default implementation for legacy solvers provided
+   * \param JointModelGroup - the planning group being proposed to be solved by this IK solver
+   * \return a string specifying the reason the joint model group is not supported, or an empty string if supported
+   */
+  virtual const std::string supportsGroup(const moveit::core::JointModelGroup *jmg) const;
 
   /**
    * @brief  Set the search discretization
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -180,8 +180,8 @@ public:
                                 const kinematics::KinematicsQueryOptions &options = kinematics::KinematicsQueryOptions()) const = 0;
 
   /**
-   * @brief Given a set of desired poses for a planning group with multiple end-effectors, search for the joint angles 
-   * required to reach it them. This is useful for e.g. biped robots that need to perform whole-body IK. 
+   * @brief Given a set of desired poses for a planning group with multiple end-effectors, search for the joint angles
+   * required to reach them. This is useful for e.g. biped robots that need to perform whole-body IK.
    * Not necessary for most robots that have kinematic chains.
    * This particular method is intended for "searching" for a solutions by stepping through the redundancy
    * (or other numerical routines).
@@ -208,7 +208,7 @@ public:
     // For IK solvers that do not support multiple poses, fall back to single pose call
     if (ik_poses.size() == 1)
     {
-      return searchPositionIK(ik_poses[0], 
+      return searchPositionIK(ik_poses[0],
         ik_seed_state,
         timeout,
         consistency_limits,
@@ -236,7 +236,8 @@ public:
 
   /**
    * @brief Set the parameters for the solver, for use with kinematic chain IK solvers
-   * @param robot_description This parameter can be used as an identifier for the robot kinematics is computed for; For example, rhe name of the ROS parameter that contains the robot description;
+   * @param robot_description This parameter can be used as an identifier for the robot kinematics it is computed for;
+   * For example, the name of the ROS parameter that contains the robot description;
    * @param group_name The group for which this solver is being configured
    * @param base_frame The base frame in which all input poses are expected.
    * This may (or may not) be the root frame of the chain that the solver operates on
@@ -251,7 +252,8 @@ public:
 
   /**
    * @brief Set the parameters for the solver, for use with non-chain IK solvers
-   * @param robot_description This parameter can be used as an identifier for the robot kinematics is computed for; For example, rhe name of the ROS parameter that contains the robot description;
+   * @param robot_description This parameter can be used as an identifier for the robot kinematics it is computed for;
+   * For example, the name of the ROS parameter that contains the robot description;
    * @param group_name The group for which this solver is being configured
    * @param base_frame The base frame in which all input poses are expected.
    * This may (or may not) be the root frame of the chain that the solver operates on
@@ -266,7 +268,8 @@ public:
 
   /**
    * @brief  Initialization function for the kinematics, for use with kinematic chain IK solvers
-   * @param robot_description This parameter can be used as an identifier for the robot kinematics is computed for; For example, rhe name of the ROS parameter that contains the robot description;
+   * @param robot_description This parameter can be used as an identifier for the robot kinematics it is computed for;
+   * For example, the name of the ROS parameter that contains the robot description;
    * @param group_name The group for which this solver is being configured
    * @param base_frame The base frame in which all input poses are expected.
    * This may (or may not) be the root frame of the chain that the solver operates on
@@ -282,7 +285,8 @@ public:
 
   /**
    * @brief  Initialization function for the kinematics, for use with non-chain IK solvers
-   * @param robot_description This parameter can be used as an identifier for the robot kinematics is computed for; For example, rhe name of the ROS parameter that contains the robot description;
+   * @param robot_description This parameter can be used as an identifier for the robot kinematics is computed for;
+   * For example, rhe name of the ROS parameter that contains the robot description;
    * @param group_name The group for which this solver is being configured
    * @param base_frame The base frame in which all input poses are expected.
    * This may (or may not) be the root frame of the chain that the solver operates on
@@ -320,7 +324,8 @@ public:
   }
 
   /**
-   * @brief  Return the name of the frame in which the solver is operating. This is usually a link name. No namespacing (e.g., no "/" prefix) should be used.
+   * @brief  Return the name of the frame in which the solver is operating. This is usually a link name.
+   * No namespacing (e.g., no "/" prefix) should be used.
    * @return The string name of the frame in which the solver is operating
    */
   virtual const std::string& getBaseFrame() const
@@ -329,7 +334,8 @@ public:
   }
 
   /**
-   * @brief  Return the name of the tip frame of the chain on which the solver is operating. This is usually a link name. No namespacing (e.g., no "/" prefix) should be used.
+   * @brief  Return the name of the tip frame of the chain on which the solver is operating. This is usually a link name.
+   * No namespacing (e.g., no "/" prefix) should be used.
    * Deprecated in favor of getTipFrames(), but will remain for foreseeable future for backwards compatibility
    * @return The string name of the tip frame of the chain on which the solver is operating
    */
@@ -342,7 +348,7 @@ public:
   }
 
   /**
-   * @brief  Return the names of the tip frames of the kinematic tree on which the solver is operating. 
+   * @brief  Return the names of the tip frames of the kinematic tree on which the solver is operating.
    * This is usually a link name. No namespacing (e.g., no "/" prefix) should be used.
    * @return The vector of names of the tip frames of the kinematic tree on which the solver is operating
    */
@@ -352,7 +358,8 @@ public:
   }
 
   /**
-   * @brief Set a set of redundant joints for the kinematics solver to use.  This can fail, depending on the IK solver and choice of redundant joints!
+   * @brief Set a set of redundant joints for the kinematics solver to use.
+   * This can fail, depending on the IK solver and choice of redundant joints!
    * @param redundant_joint_indices The set of redundant joint indices (corresponding to
    * the list of joints you get from getJointNames()).
    * @return False if any of the input joint indices are invalid (exceed number of
@@ -410,7 +417,8 @@ public:
     default_timeout_ = timeout;
   }
 
-  /** @brief For functions that require a timeout specified but one is not specified using arguments, this default timeout is used */
+  /** @brief For functions that require a timeout specified but one is not specified using arguments,
+      this default timeout is used */
   double getDefaultTimeout() const
   {
     return default_timeout_;
@@ -424,7 +432,8 @@ public:
   KinematicsBase() :
     search_discretization_(DEFAULT_SEARCH_DISCRETIZATION),
     default_timeout_(DEFAULT_TIMEOUT),
-    tip_frame_("DEPRECATED") // help users understand why this variable might not be set (if multiple tip frames provided, this variable will be unset)
+    tip_frame_("DEPRECATED") // help users understand why this variable might not be set
+                             // (if multiple tip frames provided, this variable will be unset)
   {}
 
 protected:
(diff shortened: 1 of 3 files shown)


kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -179,6 +180,50 @@ class KinematicsBase
                                 const kinematics::KinematicsQueryOptions &options = kinematics::KinematicsQueryOptions()) const = 0;
 
   /**
+   * @brief Given a set of desired poses for a planning group with multiple end-effectors, search for the joint angles 
+   * required to reach it them. This is useful for e.g. biped robots that need to perform whole-body IK.
```
acornacorn: Typo.  Also try to limit lines to 120 chars max.

davetcoleman: I don't see any misspellings here? 
The lines I actually edited from the original function descriptions are within 120 characters I believe

kinematics_base/src/kinematics_base.cpp
```diff
@@ -48,10 +48,29 @@ void kinematics::KinematicsBase::setValues(const std::string& robot_description,
...
+  robot_description_ = robot_description;
+  group_name_ = group_name;
+  base_frame_ = removeSlash(base_frame);
+  search_discretization_ = search_discretization;
+
+  // Copy tip frames to local vector after stripping slashes
+  for (std::size_t i = 0; i < tip_frames.size(); ++i)
+    tip_frames_.push_back(removeSlash(tip_frames[i]));
+
+  // Note that we do not set tip_frame_ variable here, because it should never be used in combination with multiple tip frames
```
acornacorn: Should probably set tip_frame_ here if tip_frames.size()==1.  (Eventually perhaps we can get rid of the tip_frame_ variable and then this won't matter.)

kinematics_base/src/kinematics_base.cpp
```diff
@@ -48,10 +48,29 @@ void kinematics::KinematicsBase::setValues(const std::string& robot_description,
...
+                       const std::string& group_name,
+                       const std::string& base_frame,
+                       const std::vector<std::string>& tip_frames,
+                       double search_discretization)
+{
+  robot_description_ = robot_description;
+  group_name_ = group_name;
+  base_frame_ = removeSlash(base_frame);
+  search_discretization_ = search_discretization;
+
```
acornacorn, line 67: Should probably tip_frames_.clear() here.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -205,7 +250,22 @@ class KinematicsBase
                          double search_discretization);
 
   /**
-   * @brief  Initialization function for the kinematics
+   * @brief Set the parameters for the solver, for use with non-chain IK solvers
+   * @param robot_description This parameter can be used as an identifier for the robot kinematics is computed for; For example, rhe name of the ROS parameter that contains the robot description;
```
acornacorn: typo

acornacorn:
Looks OK.  See comments inline.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
acornacorn: What does "required to reach it them" mean?

davetcoleman: man it was late, thanks

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman: fixed, though this typo is was already there!

kinematics_base/src/kinematics_base.cpp
davetcoleman: good point

davetcoleman:
I fixed all your feedback comments, thanks for the review.

davetcoleman:
What is the status of this PR? I have future pull requests depending on this one.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -221,6 +284,37 @@ class KinematicsBase
...
+    if (tip_frames.size() == 1)
+    {
+      return initialize(robot_description,
+        group_name,
+        base_frame,
+        tip_frames[0],
+        search_discretization);
+    }
+
+    logError("moveit.kinematics_base: This kinematic solver does not support initialization with more than one tip frames");
```
isucan, line 332: I think instead of the error here you can just always set tip_frames_ to the array specified in the argument and not fail.

isucan:
+1
Maybe @sachinchitta can also comment on this?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman, line 332: This is the default implementation for `initialize()` with multiple tip frames, and it is intended to be overridden by the inheriting class if that class supports multiple tip frames. This error message exists for backwards compatibility, to allow older IK solvers like IKFast to be able to accept an `initialize()` function with an array of tips, so long as that array contains only one tip entry.

davetcoleman, line 332: I can add a comment about this in the code if you agree.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -40,6 +40,7 @@
 #include <geometry_msgs/PoseStamped.h>
 #include <moveit_msgs/MoveItErrorCodes.h>
 #include <boost/function.hpp>
+#include <console_bridge/console.h>
```
hersh, line 43: Why is this console_bridge/console.h include needed?

hersh:
This does look promising, but we are thinking to do a moveit release of existing changes first and then merge this new stuff afterwards, so it has more soak time before release.

In the meantime, it would be great to see your code that will _use_ these API changes, and to see any changes this requires in existing kinematics implementations (should be none).

Do your other changes change RobotState, for instance?  If so, you should probably roll those changes into this pull request so we can see both ends of the change at once.

robot_state/src/conversions.cpp
```diff
@@ -360,16 +360,6 @@ static bool _robotStateMsgToRobotStateHelper(const Transforms *tf, const moveit_
...
-
-    // Create error message
-    if (!missing.empty())
-    {
-      std::stringstream missing_frames;
-      std::copy(missing.begin(), missing.end(), std::ostream_iterator<std::string>(missing_frames, ", "));
-      logError("robot_state: State message is missing required frames: [%s]", missing_frames.str().c_str());
-      return false;
-    }
-
```
hersh: Why does this check for missing.empty() need to be removed?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman, line 43: I added several `logError` messages to alert user when kinematic base is attempted to be used as a multi or single tip-frame IK solver when it shouldn't be. See use cases below in this PR.

kinematics_base/src/kinematics_base.cpp
davetcoleman, line 67: Fixed.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
hersh, line 43: OK.

robot_state/src/conversions.cpp
davetcoleman: I'm not sure why this appeared in this PR, but has to do with the more recent PR here: https://github.com/ros-planning/moveit_core/pull/152. I'll try to sync branches now.

davetcoleman: Fixed

davetcoleman:
I'm in no hurry to _release_ these changes, but it would be nice to get it merged before it gets forgotten about.

Code that uses API: this is the PR for moveit_ros: [Allow planning groups to have more than one tip](https://github.com/ros-planning/moveit_ros/pull/431). Unfortunately the IK solver we are using with this is written in LISP and uses some non-public robot models of HRP2 (JSK Lab at UTokyo). I'll work on getting a public demo released. 

I do have some changes to RobotState and JointModelGroup that I forgot to PR, will do now.

davetcoleman:
> This does look promising, but we are thinking to do a moveit release of existing changes first and then merge this new stuff afterwards, so it has more soak time before release.

Ping :) What is the timeline on this now - have those changes been released?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -179,6 +188,50 @@ class KinematicsBase
...
+   * required to reach them. This is useful for e.g. biped robots that need to perform whole-body IK.
+   * Not necessary for most robots that have kinematic chains.
+   * This particular method is intended for "searching" for a solutions by stepping through the redundancy
+   * (or other numerical routines).
+   * @param ik_poses the desired pose of each tip link
+   * @param ik_seed_state an initial guess solution for the inverse kinematics
+   * @param timeout The amount of time (in seconds) available to the solver
+   * @param consistency_limits the distance that any joint in the solution can be from the corresponding joints in the current seed state
+   * @param solution the solution vector
+   * @param desired_pose_callback A callback function for the desired link pose - could be used, e.g. to check for collisions for the end-effector
```
hersh: This line needs to go away, there is no such parameter in this function.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -179,6 +188,50 @@ class KinematicsBase
...
+   * @param solution_callback A callback solution for the IK solution
+   * @param error_code an error code that encodes the reason for failure or success
+   * @param lock_redundant_joints if setRedundantJoints() was previously called, keep the values of the joints marked as redundant the same as in the seed
```
hersh, line 201: lock_redundant_joints does not match any parameters in the function.

The KinematicsQueryOptions parameter is not documented.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
hersh, line 332: I think this is fine.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
hersh, line 201: Looking at the original file, I see that these same function parameter comment problems exist for all the functions.  I'll merge this now and fix those in a separate pull request.

kinematics_base/src/kinematics_base.cpp
```diff
@@ -83,3 +106,14 @@ bool kinematics::KinematicsBase::setRedundantJoints(const std::vector<std::strin
 {
   return (!str.empty() && str[0] == '/') ? removeSlash(str.substr(1)) : str;
 }
+
+const std::string kinematics::KinematicsBase::supportsGroup(const moveit::core::JointModelGroup *jmg) const
```
hersh: I did a pull request against your branch with a change to this API.  I didn't change the stuff that calls it though, that still needs to be done.

hersh:
Once the supportsGroup() call is changed like so: https://github.com/davetcoleman/moveit_core/pull/3 I am ready to merge this.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -179,6 +188,50 @@ class KinematicsBase
...
+  {
+    // For IK solvers that do not support multiple poses, fall back to single pose call
+    if (ik_poses.size() == 1)
+    {
+      return searchPositionIK(ik_poses[0],
+        ik_seed_state,
+        timeout,
+        consistency_limits,
+        solution,
+        solution_callback,
```
hersh: This should not always call the version of searchPositionIK() which takes a solution_callback function.  Instead it should check if solution_callback is defined; if so, call the version with, if not, call the version without.  This fixes a "todo" in pull request #155.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman: Done. Note: those incorrect comments were already there / are in hydro-devel

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman, line 201: Ok, I removed them then just added them back when I saw your second comment.

kinematics_base/src/kinematics_base.cpp
davetcoleman: I have not applied this function anywhere else in MoveIt! yet, I'm waiting on this PR to be approved first.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman: Good catch, thanks. Fixed.

davetcoleman:
As much as I want to see this merged, I also want to point out that this plugin interface would be a good candidate for a complete overhaul in order to improve performance and reduce function heading duplication. This might be a good time to address these issues - I've heard some others with similar thoughts. I think we could still do this while maintaining backwards compatibility. 

**Improvements**
- Pass a read-only reference to an already-loaded RobotModel to significantly reduce loading time. Currently the URDF is generally loaded for _each_  planning group that has an IK solver, as well as many other places throughout MoveIt!
- In the `initialize()` function, also pass references to the planning group instead of a string
- Fix the [IKCallbackFn](https://github.com/davetcoleman/moveit_core/blob/kinematic_base/kinematics_base/include/moveit/kinematics_base/kinematics_base.h#L87) so that it actually works correctly with MoveIt instead of needing a function adapter in [ikCallbackFnAdapter](https://github.com/ros-planning/moveit_core/blob/hydro-devel/robot_state/src/robot_state.cpp#L1192)
- Depreciate all the older functions, reducing to just one `getPositionIK` heading as discussed [here](https://github.com/ros-planning/moveit_core/issues/153)
- In this new unified IK function, pass a `robot_state` pointer instead of a `geometry_msgs::Pose` and `std::vector<double> ik_seed_state`. moveit_core is suppose to be ROS-agnostic anyway.

I do realize, however, this might cause some dependency issues and perhaps this is too difficult.

hersh:
The complete overhaul which you mention is something we've been wanting to do for a while now, but haven't had time to work on.  Some of the delays you saw trying to get your stuff merged was due to us thinking about the complete-overhaul option.

I also am frustrated by the complexity of the code that loads kinematic solvers.  I have written plugin loaders before for RViz and it can be much simpler, it seems to me.

You could create a github "issue" called "improve KinematicsBase API" or something and list your points.  Then we can discuss further actions there.

I don't think we should wait to merge these changes until the big overhaul is done though.  Let's get these in and see how we like them, and let their various features inform the new API when we have time to write it.

sachinchitta:
There is a big list of changes (which we have been keeping track of) that needs to go into the kinematics plugins as hersh mentioned. It is not something we will do lightly though since it will touch every aspect of MoveIt! and will require significant re-factoring to preserve backwards compatibility. As hersh suggests, lets get this group of changes in and we can prototype other stuff later.

davetcoleman:
Agreed, I'd really like to get these current PRs merged first. Issue moved to [here](https://github.com/ros-planning/moveit_core/issues/161).

davetcoleman:
https://github.com/davetcoleman/moveit_core/pull/3 has been merged. This should be ready to merge as soon as Travis finishes.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -179,6 +186,63 @@ class KinematicsBase
...
+   * @param ik_seed_state an initial guess solution for the inverse kinematics
+   * @param timeout The amount of time (in seconds) available to the solver
+   * @param consistency_limits the distance that any joint in the solution can be from the corresponding joints in the current seed state
+   * @param solution the solution vector
+   * @param solution_callback A callback solution for the IK solution
+   * @param error_code an error code that encodes the reason for failure or success
+   * @param lock_redundant_joints if setRedundantJoints() was previously called, keep the values of the joints marked as redundant the same as in the seed
+   * @return True if a valid solution was found, false otherwise
+   */
+  virtual bool searchPositionIK(const std::vector<geometry_msgs::Pose> &ik_poses,
```
Igorec, line 204: It would be nice to have index of the actually moving tip in the parameter list.
Like this:

``` c++
virtual bool searchPositionIK(const std::vector<geometry_msgs::Pose> &ik_poses,
                              int search_tip_index,
                              const std::vector<double> &ik_seed_state,
                              ...);
```

hersh, line 204: @davetcoleman can correct me, but I believe the idea is that all the tips in the ik_poses vector are supposed to be moving, not just one.  The size and indexing of the ik_poses array should match that of the tip_frames vector sent to initialize() or setValues().  If there are more "tips" in the group than in tip_frames, the solver might be programmed to keep those other tips at their same poses.

davetcoleman, line 204: @hersh is correct, the idea is that you can have more than one tip (or end effector) moving at the same time and your IK solver needs to be able to solve for all of them at once.

Igorec, line 204: But now function `searchPositionIK` is called from`moveit :: core :: RobotState :: setFromIK` and only one "tip" at the input does not correspond `ik_seed_state`. Am I wrong? In the future this will change?

davetcoleman, line 204: [This pull request](https://github.com/ros-planning/moveit_core/pull/155/files#diff-54f29b1e30179e9b1c3808cb99098868R1392) will be merged once KinematicBase is modified, and will fix this issue.

Igorec, line 204: Maybe it is incorrect, but I want to use this functionality to control collaborative manipulators. I move one 'tip' and others move together, so as not to drop the payload.

davetcoleman, line 204: I believe this new functionality will enable you to do that - your 'master' tip will be used to calculate the 'slave' tip pose (using reference frame conversions), then you will send both of these poses to the IK solver and it will return the valid joint positions for this non-kinematic chain

Igorec, line 204: Ok, thank you.

davetcoleman:
This PR has been open for a month and a half, can we merge it now? I'm afraid I'm going to start forgetting/lose motivation to address all the downstream PRs that are waiting on this one.

isucan:
ping. @sachinchitta -- do we have something that will be broken by merging this?

sachinchitta:
I am currently testing this PR combined with the one in moveit_ros. There are a couple of tests that I rely on for kinematics - they are in moveit_pr2 (pr2_kinematics_tests/kinematics). With these two PRs, the tests are currently failing in the initialize step. You can duplicate this by running:
rostest pr2_plugin_test.launch in that directory.

The other test to try is kdl_plugin_test.launch. Let me know when they work and I will look again into merging these two PRs.

davetcoleman:
Please look into merging this again, per: https://github.com/ros-planning/moveit_ros/issues/450

hersh:
I tested this and #155 yesterday with fresh compiles of all the moveit stuff and the tests passed.

davetcoleman:
Hurray for finally merging this!

hersh:
Yeah, well, there is still some issue with building it.  Somehow when building the tests, the /opt/ros/hydro version of kinematics_base.h is used.  Only when I deleted that .h file out of /opt/ros/hydro did it compile correctly.  I think for a release build, it would be fine since everything is built on a clean system.  But for most of us doing a "git pull" from a catkin workspace, it will cause problems.

davetcoleman:
I don't understand why it is doing that, but that is probably an issue greater than just this pull request, right? Also, once we release an updated MoveIt to debian that will go away.

hersh:
I'm working on it now.  Seems like a side-effect of the order of include_directories() calls in pr2_moveit_tests/CMakeLists.txt.

PR #151, opened 2014-02-25
Constraint samplers bug fix and improvements

davetcoleman:
As I'm implementing my own constraint sampler I've found the following issues:
- In the default JointConstraintSampler, if you have any unbound joints there is a bug where it accesses memory outside of it array. The mistake was that it was get the `getFirstVariableIndex()` of the joint with respect to the _robot state_ instead of the _joint model group_
- I fixed a bug where the default JointConstraintSampler will allow a planner to sample states even when it has not been properly configured. Both the default JointConstraintSampler and IKConstraintSampler use the `is_valid_` boolean to determine if the sampler has properly been setup, yet the constructor of the default ConstraintSampler set this flag to true under only one condition - that a valid joint model group was passed in. It should not do this, but instead wait until `configure()` has succeeded, otherwise a segfault will occur when `sample()` is called. I also clarified the documentation of this flag
- I added some comments I used when reading through the default constraint sampler logic

### DIFF
--- a/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
+++ b/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
@@ -235,7 +235,9 @@ public:
                        unsigned int max_attempts) = 0;
 
   /**
-   * \brief Returns whether or not the constraint sampler is valid or not.  To be valid, the joint model group must be available in the kinematic model.
+   * \brief Returns whether or not the constraint sampler is valid or not.
+   * To be valid, the joint model group must be available in the kinematic model and configure() must have successfully
+   * been called
    *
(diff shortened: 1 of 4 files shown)


isucan:
+100 on the joint index bug Dave! Thanks.

skohlbr:
Sounds like this could fix the issues I experienced preceding DRC Trials: [OMPL planning pipeline crashes when using path constraints](https://groups.google.com/forum/#!topic/moveit-users/FLIkvFAjQT4)

Also, could fix ros-planning/moveit_ros#366

davetcoleman:
Yes, the bug occurred when the JointConstraintSampler was used without specifying joint constraints for _every_ joint. In that case, there existed some unconstrained joints that had an indexing bug.

PR #154, opened 2014-02-27
update how we find eigen: this is needed for indigo

isucan:


davetcoleman:
This is needed for a new version of Ubuntu or just Indigo?

isucan:
Just indigo

PR #155, opened 2014-02-28
Allow joint model group to have use IK solvers with multiple tip frames

davetcoleman:
**Note:** _This commit will not build until the pending PR for moveit_core is merged ros-planning/moveit_core#149_

Adds functionality to support multiple tip-frames (non-chain kinematics) in IK solutions. 
- For `JointModelGroup`, changes `canSetStateFromIK` to check all available tip frames
- In `RobotState`, changes `setFromIK` to send a pose for all necessary tip frames that an IK solver requires. For instance, if your tree-kinematics IK solver supports 4 tips (4 end effectors, such as two hands and two feet), it will send an IK request to `KinematicsBase` with 4 poses.

### DIFF
--- a/robot_model/src/joint_model_group.cpp
+++ b/robot_model/src/joint_model_group.cpp
@@ -554,29 +554,49 @@ bool moveit::core::JointModelGroup::canSetStateFromIK(const std::string &tip) co
 {
   const kinematics::KinematicsBaseConstPtr& solver = getSolverInstance();
   if (!solver || tip.empty())
-    return false;
-  const std::string &tip_frame = solver->getTipFrame();
-  if (tip_frame.empty())
-    return false;
-  
-  // remove frame reference, if specified
-  const std::string &tip_local = tip[0] == '/' ? tip.substr(1) : tip;
-  const std::string &tip_frame_local = tip_frame[0] == '/' ? tip_frame.substr(1) : tip_frame;
-  
-  if (tip_local != tip_frame_local)
   {
-    if (hasLinkModel(tip_frame_local))
-    {
-      const LinkModel *lm = getLinkModel(tip_frame_local);
-      const LinkTransformMap &fixed_links = lm->getAssociatedFixedTransforms();
-      for (LinkTransformMap::const_iterator it = fixed_links.begin() ; it != fixed_links.end() ; ++it)
-        if (it->first->getName() == tip_local)
-          return true;
-    }
+    logError("Cannot set state, no solver or tip is empty");
     return false;
   }
-  else
-    return true;
+
+  const std::vector<std::string> &tip_frames = solver->getTipFrames();
+
+  if (tip_frames.empty())
+  {
+    logDebug("Group %s has no tip frame(s)", name_.c_str());
+    return false;
+  }
+
+  // loop through all tip frames supported by the JMG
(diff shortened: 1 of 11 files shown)


robot_model/src/joint_model_group.cpp
```diff
@@ -554,29 +554,49 @@ bool moveit::core::JointModelGroup::canSetStateFromIK(const std::string &tip) co
 {
   const kinematics::KinematicsBaseConstPtr& solver = getSolverInstance();
   if (!solver || tip.empty())
+  {
+    logError("Cannot set state, no solver or tip is empty");
```
hersh: These logError() printouts need to be removed.  The point of the function is for the caller to know if setFromIK() is going to work, so the caller is doing the right thing, even if the return value is false.

robot_model/src/joint_model_group.cpp
```diff
@@ -554,29 +554,49 @@ bool moveit::core::JointModelGroup::canSetStateFromIK(const std::string &tip) co
...
     return false;
-  const std::string &tip_frame = solver->getTipFrame();
-  if (tip_frame.empty())
+  }
+
+  const std::vector<std::string> &tip_frames = solver->getTipFrames();
+
+  if (tip_frames.empty())
+  {
+    logDebug("Group %s has no tip frame(s)", name_.c_str());
```
hersh, line 563: This logError() needs to be removed also.

robot_state/src/robot_state.cpp
```diff
@@ -1332,9 +1389,8 @@ bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const Eigen
     // compute the IK solution
     std::vector<double> ik_sol;
     moveit_msgs::MoveItErrorCodes error;
-    if (ik_callback_fn ?
-        solver->searchPositionIK(ik_query, seed, timeout, consistency_limits, ik_sol, ik_callback_fn, error, options) :
-        solver->searchPositionIK(ik_query, seed, timeout, consistency_limits, ik_sol, error, options))
+    // \todo no function call that accepts multiple poses and no callback has been created. is it really needed?
```
hersh: This "todo" can be done (and then removed from this pull request please) by changing the default implementation of KinematicsBase::searchPositionIK() which takes multiple tip frames to check if the callback function is NULL and if so, call the searchPositionIK() without a callback argument.  See https://github.com/ros-planning/moveit_core/pull/149/files#r10730403.

davetcoleman: Fixed, thanks.

robot_model/src/joint_model_group.cpp
davetcoleman, line 563: You mean `logDebug()`? I think its useful feedback to aid the user in understanding why their planning group is unable to use an IK solver. This area is currently kind of difficult IMHO to debug.

robot_model/src/joint_model_group.cpp
davetcoleman: Removed.

robot_model/src/joint_model_group.cpp
hersh, line 563: My mistake, logDebug() is fine to leave in.

PR #157, opened 2014-02-28
generate headers in devel space instead of build space

dirk-thomas:
With the upcoming catkin version passing absolute INCLUDE_DIRS to catkin_package() will be allowed (ros/catkin#600). But since your absolute _build_ space path is temporary downstream packages will fail to find that folder. (Before any absolute path was ignored but there where use cases where it should be possible.)

Instead this pull request generates the header in the _devel_ space which catkin is aware of and assumes that these generated files will be installed by the user (similar as generated message headers).

Please review and test - I haven't even tried to compile the package after the change :wink: This fix should be released in _Groovy_, _Hydro_ and _Indigo_.


isucan:
Vincent did this better ;)

dirk-thomas:
As far as I can see from his patch (8a76161c4bef8b0dd85eea94f642549788d123a6) the folder where the header is generated to is never being added to the `include_directories()`. My guess would be that it will fail to build when you try an isolated build (without another package e.g. containing messages which will add the devel include folder for you). I guess the farm will tell us when it tries to build the Debian package.

PR #159, opened 2014-03-03
avoid unnecessary calculations

acornacorn:
If we are not going to use the "missing" vector then we should not create it
(avoid an expensive operation).

This is related to the changes in https://github.com/ros-planning/moveit_core/pull/152 and https://github.com/ros-planning/moveit_core/pull/148

### DIFF
--- a/robot_state/src/conversions.cpp
+++ b/robot_state/src/conversions.cpp
@@ -51,7 +51,7 @@ namespace core
 namespace
 {
 
-static bool _jointStateToRobotState(const sensor_msgs::JointState &joint_state, RobotState& state, std::set<std::string> *missing)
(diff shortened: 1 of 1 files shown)


isucan:
+1

davetcoleman:
+1 good call

davetcoleman:
Concerning the newly defined PR policy - I thoroughly read through the changes in this PR and am experienced with it since I made related changes to it very recently, Ioan also +1, so it is ok that I merged. Correct?

isucan:
If you are confident and understand the changes you merge,  it is fine to merge. The way I like to think about it is that you share responsibility with the submitter in fixing the problem if you merge something bad.

PR #160, opened 2014-03-08
computeCartesianPath waypoints double-up fix

mszarski:
computeCartesianPath appends full trajectories between waypoints when given a vector of waypoints. As trajectories include their endpoints, this leads to the combined trajectory being generated with duplicate points at waypoints, which can lead to pauses or stuttering.

This change skips the first point in trajectories generated between waypoints.

### DIFF
--- a/robot_state/src/robot_state.cpp
+++ b/robot_state/src/robot_state.cpp
@@ -1661,12 +1661,18 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup *gro
     if (fabs(wp_percentage_solved - 1.0) < std::numeric_limits<double>::epsilon())
     {
       percentage_solved = (double)(i + 1) / (double)waypoints.size();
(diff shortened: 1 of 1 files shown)


robot_state/src/robot_state.cpp
```diff
@@ -1661,12 +1661,18 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup *gro
     if (fabs(wp_percentage_solved - 1.0) < std::numeric_limits<double>::epsilon())
     {
       percentage_solved = (double)(i + 1) / (double)waypoints.size();
-      traj.insert(traj.end(), waypoint_traj.begin(), waypoint_traj.end());
+      std::vector<RobotStatePtr>::iterator start = waypoint_traj.begin();
+      if(i > 0)
```
isucan: maybe add && !waypoint_traj.empty() ?

isucan:
This looks good, but I did not thoroughly check this.

PR #164, opened 2014-04-22
Manipulability Index Error for few DOF

costashatz:
When the group has fewer than 6 DOF, the Jacobian is of the form 6xM and when multiplied by its transpose, forms a 6x6 matrix that is singular and its determinant is always 0 (or NAN if the solver cannot calculate it).

Since calculating the SVD of a Jacobian is a costly operation, I propose to retain the calculation of the Manipulability Index through the determinant for 6 or more DOF (where it produces the correct result), but use the product of the singular values of the Jacobian for fewer DOF.

### DIFF
--- a/kinematics_metrics/src/kinematics_metrics.cpp
+++ b/kinematics_metrics/src/kinematics_metrics.cpp
@@ -114,16 +114,48 @@ bool KinematicsMetrics::getManipulabilityIndex(const robot_state::RobotState &st
   double penalty = getJointLimitsPenalty(state, joint_model_group);
   if (translation)
   {
-    Eigen::MatrixXd jacobian_2 = jacobian.topLeftCorner(3,jacobian.cols());
-    Eigen::MatrixXd matrix = jacobian_2*jacobian_2.transpose();
-    // Get manipulability index
-    manipulability_index = penalty * sqrt(matrix.determinant());
+    if(jacobian.cols()<6)
+    {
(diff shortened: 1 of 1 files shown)


davetcoleman:
I sort of understand the math behind this, but not enough to confidently merge this. I see this PR does not change anything for robots with 6 or more DOF (which is fine), so it will only affect a small number of MoveIt! robots that have fewer DOFs. Certainly the author @sachinchitta would be best suited to merge this

I'm interested though - could you send me a link to an article explaining this technique for calculating the manipulability index?

progtologist:
In most books and papers the Manipulability measure is introduced as det(J*J^{T}) for redundant manipulators and as det(J) for non-redundant manipulators.

Keeping the notation found in "Robotics, Modeling and Control"  by Bruno Siciliano, Lorenzo Sciavicco, Luigi Villani, Giuseppe Oriolo in Chapter 3, pages 153-156 when the manipulator in question has fewer than 6DoF then the workspace of that manipulator is not se(3), but a subspace where r = n.

In that case the Manipulability measure is calculated by det(J). If the Jacobian is arbitrary sized (RxN dimensions), det(J) = det(J \* J^{T}), but in ROS the Jacobian matrix is always a (6xN) which causes the NaN values, as the extra rows are singular.

By using the product of the singular values - although not computationally efficient - we overcome this problem.

I hope I have helped in making this clear.

sachinchitta:
Looks like a good update - could you please open a corresponding request against indigo-devel as well?

PR #166, opened 2014-04-30
Fix compiler warnings

adolfo-rt:
Similar to ros-planning/moveit_ros#451, but targeting moveit_core.

Most fixes are on enforcing compiler strictness and minor performance improvements. However, 4130e2b is behavior-changing. I believe it's a bugfix.


davetcoleman:
+1 visual review

isucan:
+1

PR #168, opened 2014-05-01
Perform self-collision checks with padded robot.

adolfo-rt:
Self-collision checking is currently done with an unpadded robot. My expectation is that if a robot has a padding specification, it should not only be used for environment collision checks, but also for self-collision checks.

I've tested this patch on my end and it works, but I don't know to which extent it might be affecting code that actually expects self-collision checks to be unpadded.

See also @k-okada's comment on https://github.com/ros-planning/moveit_ros/issues/341#issuecomment-32701710.

### DIFF
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -523,8 +523,8 @@ void planning_scene::PlanningScene::checkCollision(const collision_detection::Co
 
   if (!res.collision || (req.contacts && res.contacts.size() < req.max_contacts))
   {
-    // do self-collision checking with the unpadded version of the robot
-    getCollisionRobotUnpadded()->checkSelfCollision(req, res, kstate, getAllowedCollisionMatrix());
+    // do self-collision checking with the padded version of the robot
+    getCollisionRobot()->checkSelfCollision(req, res, kstate, getAllowedCollisionMatrix());
   }
 }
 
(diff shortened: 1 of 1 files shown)


k-okada:
very nice!!

isucan:
-1 
One uses padding for things like collision avoidance with the intention of maintaining higher clearance when there are calibration issues or pose estimation inaccuracies. In addition, when doing things like self filtering, again, we want some padding.

For self collision detection, we may want to use a separate set of padding values (currently all 0 -- the unpadded robot), but I do not think they should be the same as in the other situations.
So, if you need it, I think we can definitely use a differently padded version of the robot, but not the same one.

skohlbr:
For most robot systems, uncertainty/noise in sensing of the environment is much higher than in sensing internal state (e.g. joint position values). For this reason, I'd argue that using the unpadded robot makes sense for self-collision checks, while using the padded robot for environment collision checks (essentially adding some buffer to account for noise in environment sensing). 

/edit: Ioan beat me to it, same reason :)

mikeferguson:
-1 for the same reasons as @isucan

adolfo-rt:
@skohlbr wrote:

> For most robot systems, uncertainty/noise in sensing of the environment is much higher than in sensing internal state (e.g. joint position values).

True, the magnitudes are not the same. Still, I don't want motion plans potentially being an epsilon away from self-collision. I've already seen plans in which the results are within joint offset calibration / control / modeling (collision mesh accuracy) errors.

@isucan wrote:

> So, if you need it, I think we can definitely use a differently padded version of the robot, but not the same one.

This sounds like the way to go. Thanks for the feedback.

PR #170, opened 2014-05-07
IterativeParabolicTimeParameterization now ignores virtual joints.

hersh:
[edit: see next comment for updated info]

When checking if all joints are single-DOF, it accepts multi-DOF joints only if they are
also virtual.

Previously it would fail on my planned trajectories which involve only single-DOF joints plus the virtual "world joint" which I needed to include in the planning group in order to visualize the plan properly.  The virtual joints don't really exist, so I think this should be fine.

I have run this with my robot model including the world joint and it seems to work fine.

### DIFF
--- a/trajectory_processing/src/iterative_time_parameterization.cpp
+++ b/trajectory_processing/src/iterative_time_parameterization.cpp
@@ -449,25 +449,6 @@ bool IterativeParabolicTimeParameterization::computeTimeStamps(robot_trajectory:
     return false;
   }
 
-  // Collect the set of virtual joint names so we can ignore them.
-  std::set<std::string> virtual_joints;
-  const std::vector<srdf::Model::VirtualJoint> &vjoints =
-    trajectory.getRobotModel()->getSRDF()->getVirtualJoints();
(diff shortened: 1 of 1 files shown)


hersh:
After the discussion in https://github.com/ros-planning/moveit_ros/issues/453 I thought I would re-visit this pull request, since it also deals with special cases for virtual joints.

Looking through the code, I can't see any reason why IterativeParabolicTimeParameterization won't work with multi-DOF joints.  All the code internally uses variables and the indices of variables, it doesn't refer to joints as such at all.  From my previous experience allowing the "world joint" (6DOF) to pass through it seems to work fine in practice.

I realize that maybe someone would like to have a better time parameterizer that does something smarter with multi-DOF joints, but until someone writes one, I don't see any reason to prevent people from using this one with them.

Therefore in the second commit that I just added, the code that checks for multi-DOF and virtual joints has been removed entirely.

isucan:
+1

davetcoleman:
I'm happy to see it gone. Agreed we need something better in the future.

PR #171, opened 2014-05-15
Moved robot_state library code into robot_model library.

hersh:
This is a (likely) necessary precondition to writing the new inverse-kinematics API I would like to write.

Without this change my IK code would create a circular dependency between the libraries, and I think that is at least a bad idea and maybe doesn't work at all.

Currently moveit_core consists of around 20 tiny separate libraries, each of which depend on several others in a complicated dependency graph.  External users of these libraries seem to just use the catkin build machinery to say "I need moveit_core" which gets them the whole set of libraries.  So no one outside of moveit_core actually cares about library names.

I think it may be a good idea to combine many more and simplify the build structure, but for now I'm just doing the one I need.


isucan:
We worked hard to actually make the separation happen. Model != State. I agree that this is necessary, but I would much prefer compiling moveit_core into one library, rather than merging just these two. This said, I am fine with merging this as a first step.

acornacorn:
+1

hersh:
@isucan I'm sorry about denigrating the hard work of the library separation.  What was the use case for separating the libraries?

I understand that model and state are different concepts, and they are indeed separated nicely into different classes.  That part I definitely don't want to change.

isucan:
@hersh I don't see this as denigrating the work :) This is indeed a step forward. I was just saying, as you do as well, that we should put this all into one single .so; that is not so much fun to do though. 
I would strongly prefer to keep robot_model & robot_state headers in different folders though. I think this is important to underline the differences between these sets of headers. RobotModel is meant as a constant in the system, RobotState is meant as something relatively cheap to construct that anybody can create and modify as they see fit.

hersh:
Sigh.  I have found a much easier-to-implement solution to my kinematics API needs.  It doesn't require either of the changes in this pull request.

Closing without merging (for now).

PR #172, opened 2014-05-16
fix pick/place approach/retreat on indigo/14.04

mikeferguson:
I'm not entirely sure _why_ this is a problem, and it has no issue with a source checkout on Hydro/12.04, but as long as rotated_direction is a reference, the rotated_direction vector is always wrong with a source checkout on Indigo and Ubuntu 14.04 (all values nearly zero). Could be a change in Eigen, or the compiler I suppose. @isucan, any thoughts?


isucan:
very weird. This sounds like a memory alignment issue; also, taking const & to temporary may be a bit weird.
+1

mikeferguson:
Ping? Any thoughts on merging this?

PR #174, opened 2014-05-27
Fixed bug in RevoluteJointModel::distance() giving large negative numbers.

hersh:
RevoluteJointModel::distance() was giving large negative values for a continuous joint, and upon investigation it turns out that there is nothing to enforce the continuous joint values to be between -pi and +pi or 0 and 2_pi.  The code in distance() was correct if the values *were_ constrained, so I just added an extra "fmod()" call.


davetcoleman:
+1 makes sense to me

isucan:
+1

PR #175, opened 2014-06-24
need angles dependency for tests

sachinchitta:


isucan:
This is a test depend, so it should be defined as such in package.xml - not run/ build depend

sachinchitta:
Yup, you are right - done. Will do a new release of moveit_core tomorrow if this passes Travis.

acornacorn:
Travis failed because angles is a test-only depend but is exported as a dependent package dependency.  I think you need to remove angles from the catkin_package macro and add it as a standalone find_package (without REQUIRED).

hersh:
@sachinchitta, have you compiled this?  For some reason Travis didn't build from your more recent changes, so the latest build result I see is a failure.

sachinchitta:
Yes, I did compile it. I am not sure why Travis did not rebuild.

PR #178, opened 2014-07-02
Added getSubgroups() helper function to joint model groups

davetcoleman:
There was a `std::string getSubGroupsNames()` but no method to get the `JointModelGroup*` pointers directly.


sachinchitta:
Could you add a corresponding pull request for moveit_core indigo as well.

sachinchitta:
LGTM.

davetcoleman:
I'll do a PR to indigo once this is merged. I don't want to be making edits from feedback on two different branches.

PR #177, opened 2014-07-01
New setToRandomPositions function with custom rand num generator

davetcoleman:
Added new setToRandomPositions function that allows custom random number generator to be specified

Used for benchmarking with stochastic behavior. Useful with https://github.com/ros-planning/random_numbers/pull/7

PR #180, opened 2014-07-14
Allow setFromIK() with multiple poses to single IK solver

davetcoleman:
Allows a call to setFromIK() that has multiple tips and desired poses to send them all to the same IK solver, if that solver allows it. If no solver exists for the current planning group or that solver does not support non-chain kinematics, the previous MoveIt! behavior is maintained whereby it looks for subgroups that are chains that have their own IK solvers and calls them individually.

This is a follow up of https://github.com/ros-planning/moveit_core/pull/149
### Minor Changes

Reduced duplicate code by moving two common code snippets into their own functions:
- Created new helper function for changing a pose to the base frame of an IK solver
- Created new helper function for converting a Eigen pose to a geometry_msgs pose

Contains outstanding PR:
https://github.com/ros-planning/moveit_core/pull/177

### DIFF
--- a/robot_state/include/moveit/robot_state/robot_state.h
+++ b/robot_state/include/moveit/robot_state/robot_state.h
@@ -619,7 +619,6 @@ as the new values that correspond to the group */
    */
   bool setToIKSolverFrame(Eigen::Affine3d &pose, const kinematics::KinematicsBaseConstPtr& solver);
 
-
   /**
    * \brief Convert the frame of reference of the pose to that same frame as the IK solver expects
    * @param pose - the input to change
--- a/robot_state/include/moveit/robot_state/robot_state.h
+++ b/robot_state/include/moveit/robot_state/robot_state.h
@@ -611,6 +611,21 @@ as the new values that correspond to the group */
   /** \brief For a given group, copy the position values of the variables that make up the group into another location, in the order that the variables are found in the group. This is not necessarily a contiguous block of memory in the RobotState itself, so we copy instead of returning a pointer.*/  
   void copyJointGroupPositions(const JointModelGroup *group, Eigen::VectorXd& values) const;
 
+  /**
+   * \brief Convert the frame of reference of the pose to that same frame as the IK solver expects
+   * @param pose - the input to change
+   * @param solver - a kin solver whose base frame is important to us
+   * @return true if no error
+   */
+  bool setToIKSolverFrame(Eigen::Affine3d &pose, const kinematics::KinematicsBaseConstPtr& solver);
+
+  /**
+   * \brief Convert an Eigen pose to a geometry_msgs pose
+   * @param pose - input
+   * @param pose_msg - output
+   */
+  void poseToMsg(const Eigen::Affine3d &pose, geometry_msgs::Pose &pose_msg);
+
   /** \brief If the group this state corresponds to is a chain and a solver is available, then the joint values can be set by computing inverse kinematics.
       The pose is assumed to be in the reference frame of the kinematic model. Returns true on success.
       @param pose The pose the last link in the chain needs to achieve
@@ -710,7 +725,24 @@ as the new values that correspond to the group */
                  unsigned int attempts = 0, double timeout = 0.0,
                  const GroupStateValidityCallbackFn &constraint = GroupStateValidityCallbackFn(),
                  const kinematics::KinematicsQueryOptions &options = kinematics::KinematicsQueryOptions());
-  
+
+  /**
+      \brief setFromIK for multiple poses and tips (end effectors) when no solver exists for the jmg that can solver for
+      non-chain kinematics. In this case, we divide the group into subgroups and do IK solving individually
+      @param poses The poses the last link in each chain needs to achieve
+      @param tips The names of the frames for which IK is attempted.
+      @param consistency_limits This specifies the desired distance between the solution and the seed state
+      @param attempts The number of times IK is attempted
+      @param timeout The timeout passed to the kinematics solver on each attempt
+      @param constraint A state validity constraint to be required for IK solutions */
+  bool setFromIKSubgroups(const JointModelGroup *group,
+                 const EigenSTL::vector_Affine3d &poses,
+                 const std::vector<std::string> &tips,
+                 const std::vector<std::vector<double> > &consistency_limits,
+                 unsigned int attempts = 0, double timeout = 0.0,
+                 const GroupStateValidityCallbackFn &constraint = GroupStateValidityCallbackFn(),
+                 const kinematics::KinematicsQueryOptions &options = kinematics::KinematicsQueryOptions());
+
   /** \brief Set the joint values from a cartesian velocity applied during a time dt
    * @param group the group of joints this function operates on
    * @param twist a cartesian velocity on the 'tip' frame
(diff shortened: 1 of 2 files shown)


isucan:
@sachinchitta ? I am fine with merging this.

robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -611,6 +611,29 @@ as the new values that correspond to the group */
...
+   * @return true if no error
+   */
+  bool setToIKSolverFrame(Eigen::Affine3d &pose, const std::string &ik_frame);
+
+  /**
+   * \brief Convert an Eigen pose to a geometry_msgs pose
+   * @param pose - input
+   * @param pose_msg - output
+   */
+  void poseToMsg(const Eigen::Affine3d &pose, geometry_msgs::Pose &pose_msg);
```
sachinchitta: There's a eigen to msg conversions package that should be used - this is a duplication and should not be in robot_state.h

robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -611,6 +611,29 @@ as the new values that correspond to the group */
...
+   */
+  bool setToIKSolverFrame(Eigen::Affine3d &pose, const kinematics::KinematicsBaseConstPtr& solver);
+
+  /**
+   * \brief Convert the frame of reference of the pose to that same frame as the IK solver expects
+   * @param pose - the input to change
+   * @param ik_frame - the name of frame of reference of base of ik solver
+   * @return true if no error
+   */
+  bool setToIKSolverFrame(Eigen::Affine3d &pose, const std::string &ik_frame);
```
sachinchitta, line 628: I believe that this is already done by the transforms class (transforming a pose to a new frame). Why duplicate it here (unless its not easily available here)?

robot_state/src/robot_state.cpp
```diff
@@ -1231,146 +1225,270 @@ bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const Eigen
...
+  {
+    valid_solver = false;
+  }
+  // Check if this jmg's IK solver can handle multiple tips (non-chain solver)
+  else if (poses_in.size() > 1)
+  {
+    std::string error_msg;
+    if(!solver->supportsGroup(jmg, &error_msg))
     {
-      if (hasAttachedBody(tip))
```
sachinchitta, line 1250: Its a little hard for me to parse through these changes but this looks like support for attached bodies IK is being taken out?

robot_state/src/robot_state.cpp
```diff
@@ -1402,45 +1522,32 @@ bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const Eigen
   return false;
 }
 
-bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const EigenSTL::vector_Affine3d &poses_in, const std::vector<std::string> &tips_in,
-                                         unsigned int attempts, double timeout,
-                                         const GroupStateValidityCallbackFn &constraint, const kinematics::KinematicsQueryOptions &options)
```
sachinchitta, line 1407: Is this function being taken out of the API? Why not just add a new function and leave the old ones in to minimize API changes?

robot_state/include/moveit/robot_state/robot_state.h
davetcoleman, line 628: At the heart of this function yes, it does call `getGlobalLinkTransform()`. But it encapsulates several other lines of logic that I saw repeated throughout the `setFromIK()` functions that I thought would increase clarity by removing this sub function, including
- Checking if the two frames are already the same
- Checking if the frame is prepended with a '/' or not
- Checking if the LinkModel pointer is valid
- A line of math
- Getting the base frame from the solver

I think it should stay

robot_state/include/moveit/robot_state/robot_state.h
davetcoleman: That's a really good point. Before this PR this same code was hard-coded into robot_state inline with the code, and I didn't think about this duplication when I created this function. Removed.

robot_state/src/robot_state.cpp
davetcoleman, line 1250: Github just did a bad job of overlaying the changes I'm afraid. Both `setFromIK()` and the new `setFromIKSubgroups()` still have the `hasAttachedBody(tip)` logic, it just moved a little further down.

robot_state/src/robot_state.cpp
```diff
@@ -1231,146 +1225,270 @@ bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const Eigen
...
+      if (!tip_frame.empty() && tip_frame[0] == '/')
+        tip_frame = tip_frame.substr(1);
+
+      if (tip != tip_frame)
       {
-        const LinkModel *lm = getLinkModel((!ik_frame.empty() && ik_frame[0] == '/') ? ik_frame.substr(1) : ik_frame);
-        if (!lm)
-          return false;
-        current_pose = getGlobalLinkTransform(lm).inverse() * current_pose;
+        if (hasAttachedBody(tip))
```
davetcoleman, line 1356: This is the new location of `hasAttachedBody`

robot_state/src/robot_state.cpp
```diff
@@ -1231,146 +1225,270 @@ bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const Eigen
...
+}
 
-  // try all of the solver's possible tip frames to see if one works with our passed in pose tip frame
-  bool found_valid_frame = false;
-  for (std::vector<std::string>::const_iterator tip_frame_it = solver->getTipFrames().begin();
-       tip_frame_it < solver->getTipFrames().end(); ++tip_frame_it)
+bool moveit::core::RobotState::setFromIK(const JointModelGroup *jmg, const EigenSTL::vector_Affine3d &poses_in, const std::vector<std::string> &tips_in,
+                                         unsigned int attempts, double timeout,
+                                         const GroupStateValidityCallbackFn &constraint, const kinematics::KinematicsQueryOptions &options)
+{
```
davetcoleman, line 1251: Above is the location of the `setFromIK` function that @sachinchitta thought might have been removed. Not sure why it appears re-ordered, it is the correct order in the code.

robot_state/src/robot_state.cpp
davetcoleman, line 1407: No, it is still included but appears moved, as I noted above.

davetcoleman:
I have addressed all the concerns of @sachinchitta in https://github.com/davetcoleman/moveit_core/commit/4d5afb5584a405ea94383abc773c5b8b839c65bd, thanks for the feedback

robot_state/include/moveit/robot_state/robot_state.h
sachinchitta, line 628: Given this this uses a string frame name, it could be used to transform to any frame (not just an ik frame). Calling this function transformPoseToFrame matches better with the syntax in other places. Also, if this function is mostly used internally, maybe it should be private?

davetcoleman, line 628: I like `transformPoseToFrame()`, ok. And making it private makes sense.

davetcoleman, line 628: Actually, I was making the modifications just now, and I realized this function is really intended just for IK solvers. It first checks if the IK solver frame is the same as the robot_model frame, and if it is it skips the transform. Only if they differ does it change the frame.

If you are worried about adding too many functions, I could remove the first `setToIKSolverFrame` that takes as input the solver pointer, it isn't very useful.

PR #182, opened 2014-07-17
Added ability to return used planning context in planning pipeline

davetcoleman:
This adds a new feature to the planning request adapter chains that tracks the used planning context throughout as the various adapters call each other in succession. It allows one to pass in a context pointer and when the chain returns it keeps the planning context in existence for further usage and analysis. I'm using this to allow me to access the OMPL SimpleSetup class from within the planning pipeline, so that I can call some custom functions on it that I need.

This change also requires the planning adapters to be slightly modified to accept the context in their function headers.

Finally, I renamed two functions to improve clarity and added some inline documentation.

### DIFF
--- a/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -32,7 +32,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
-/* Author: Ioan Sucan */
+/* Author: Ioan Sucan, Dave Coleman */
 
 #ifndef MOVEIT_PLANNING_REQUEST_ADAPTER_PLANNING_REQUEST_ADAPTER_
 #define MOVEIT_PLANNING_REQUEST_ADAPTER_PLANNING_REQUEST_ADAPTER_
@@ -41,7 +41,12 @@
 #include <moveit/planning_scene/planning_scene.h>
 #include <boost/function.hpp>
 
-/** \brief Generic interface to adapting motion planning requests */
+/** \brief Generic interface to adapting motion planning requests
+ *
+ *  The PlannRequestAdapter class is both the class used for all the adapters, as well as a
+ *  custom version used to wrap a traditional motion planner into looking like an adapter
+ *
+ */
 namespace planning_request_adapter
 {
 
@@ -66,13 +71,15 @@ public:
   bool adaptAndPlan(const planning_interface::PlannerManagerPtr &planner,
                     const planning_scene::PlanningSceneConstPtr& planning_scene,
                     const planning_interface::MotionPlanRequest &req,
-                    planning_interface::MotionPlanResponse &res) const;
+                    planning_interface::MotionPlanResponse &res,
+                    planning_interface::PlanningContextPtr &context) const;
 
   bool adaptAndPlan(const planning_interface::PlannerManagerPtr &planner,
                     const planning_scene::PlanningSceneConstPtr& planning_scene,
                     const planning_interface::MotionPlanRequest &req,
                     planning_interface::MotionPlanResponse &res,
-                    std::vector<std::size_t> &added_path_index) const;
+                    std::vector<std::size_t> &added_path_index,
+                    planning_interface::PlanningContextPtr &context) const;
 
   /** \brief Adapt the planning request if needed, call the planner
       function \e planner and update the planning response if
@@ -83,7 +90,8 @@ public:
                             const planning_scene::PlanningSceneConstPtr& planning_scene,
                             const planning_interface::MotionPlanRequest &req,
                             planning_interface::MotionPlanResponse &res,
-                            std::vector<std::size_t> &added_path_index) const = 0;
+                            std::vector<std::size_t> &added_path_index,
+                            planning_interface::PlanningContextPtr &context) const = 0;
 
 };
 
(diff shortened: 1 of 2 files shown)


davetcoleman:
Travis failed because of a kdl_kinematics test, that should not be related to this PR. Any insight into why Travis says:

```
[ROSUNIT] Outputting test results to /home/travis/ros/ws_moveit/build/test_results/pr2_moveit_tests/rostest-kinematics_launch_kdl_plugin_test.xml
[ERROR] [1405572303.667538079]: The plugin failed to load. Error: Could not find library corresponding to plugin kdl_kinematics_plugin/KDLKinematicsPlugin. Make sure the plugin description XML file has the correct name of the library and that the library actually exists.
pr2_kinematics_test: /usr/include/boost/smart_ptr/shared_ptr.hpp:418: T* boost::shared_ptr<T>::operator->() const [with T = kinematics::KinematicsBase]: Assertion `px != 0' failed.
testkdl_plugin_test ... FAILURE!
FAILURE: test [kdl_plugin_test] did not generate test results
```

??

mikeferguson:
Scroll all the way up to line 3244 of the build log -- "make" failed -- in the default_planning_request_adapter_plugins -- thus I think the build failure might be related to pull request

davetcoleman:
@mikeferguson I see the problem now. This changes the plugin interface for the `planning_request_adapter_plugins` in `moveit_ros`, so this won't build on travis until both it and the moveit_ros [related PR](https://github.com/ros-planning/moveit_ros/pull/477) is also merged. Not sure how to circumvent this issue.

davetcoleman:
This can be merged despite Travis failing - it is unavoidable because Travis attempts to also build moveit_ros at the same time, which cannot be fixed until this is merged.

planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -41,7 +41,12 @@
 #include <moveit/planning_scene/planning_scene.h>
 #include <boost/function.hpp>
 
-/** \brief Generic interface to adapting motion planning requests */
+/** \brief Generic interface to adapting motion planning requests
+ *
+ *  The PlannRequestAdapter class is both the class used for all the adapters, as well as a
```
mikeferguson: spelling, PlannRequestAdapter -> PlanningRequestAdapter?

planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -153,22 +166,37 @@ bool planning_request_adapter::PlanningRequestAdapterChain::adaptAndPlan(const p
                                                                          planning_interface::MotionPlanResponse &res,
                                                                          std::vector<std::size_t> &added_path_index) const
 {
+  std::vector<std::size_t> dummy;
```
mikeferguson: This should not create dummy, it should send added_path_index on through, right?

mikeferguson:
Ok, having re-read this code a few times (and that from https://github.com/ros-planning/moveit_ros/pull/477), my sudden "oh wait" moment has been fixed by realizing this API was more complex than I thought.

+1 to the idea, I think there is a bug introduced as noted above, should pass added_path_index, not create dummy (otherwise the function is no different than the version above it). Once that is resolved, +1 to the whole PR.

planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
davetcoleman: Fixed

planning_request_adapter/src/planning_request_adapter.cpp
davetcoleman: Good question. This commit is over 3 months old, so I do not recall why I have that there or if it should be there. I've gone ahead and fixed it because I think you are right. Side rant: we really need to be more responsive in merging stuff if we want contributors to MoveIt! (and also thanks for reviewing this now, better than never)

davetcoleman:
Fixed, thanks!

@isucan I don't recall 100% what we had agreed upon about this commit, did you still want to do it differently?

isucan:
Yes, I'd like this done differently. We talked a bit about it, maybe we shold have another meeting? Or should I try to attach my suggestion here ?

mikeferguson:
@isucan, I'd appreciate if you post your comments here -- I'm interested to hear how you would do this differently.

davetcoleman:
Mostly I'm happy with my version (works fine for my research) so could just keep that in my branch and close this PR. But I'm happy to have a meeting too.

isucan:
Dave, I don't remember when we last spoke about this, but I thought this was making some OMPL-specific objects visible. It seems you expose the generic context, as I would like it to be. Is this PR still ok for you too merge as is?

isucan:
And does it require other PRs in other repos?

davetcoleman:
I am still using this PR, yes, it is ok to merge in. It exposes OMPL-specific objects via letting you get SimpleSetup in the planning context. I don't remember everything 100% because this 4 months old.

You should also merge https://github.com/ros-planning/moveit_ros/pull/477/files at the same time.

mikeferguson:
Where are we on merging this? I concur it would be useful. It doesn't appear to have anything OMPL-specific (unless you cast the returned context, but that isn't happening in this code).

mikeferguson:
@davetcoleman I think you missed reviving this one.

davetcoleman:
I'd still be interested in merging this in, btw

sachinchitta:
Hydro is EOL - please move this into jade if you still want it.

PR #188, opened 2014-07-31
Constraint sampler inverse kinematics fix

arjungm:
I added an extra argument to ConstrainedSampler::project that specifies whether or not to use the passed in state as a seed, which defaults to true.

I added this argument to all of the derived classes too.

This will let ompl_interface classes that use the ConstrainedSamplers to force randomized seeds when it detects failures at sampling goals.

The problem it fixes is talked about here
https://github.com/ros-planning/moveit_planners/issues/44#issuecomment-50715025

### DIFF
--- a/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
+++ b/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler.h
@@ -232,7 +232,7 @@ public:
    * @return True if a sample was successfully projected, false otherwise
    */
   virtual bool project(robot_state::RobotState &state,
-                       unsigned int max_attempts) = 0;
+                       unsigned int max_attempts, bool use_state_as_seed = true) = 0;
 
   /**
    * \brief Returns whether or not the constraint sampler is valid or not.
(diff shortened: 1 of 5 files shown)


davetcoleman:
This change seems harmless since it only pokes a hole to provide another setting.

However, I don't understand fully why you a goal sampler would need to change 'modes' to a random seed. Isn't a goal sampler simply finding valid states within a certain range of error? So it should already be choosing random nearby states, right? That is my understanding of this feature.

davetcoleman:
I guess I don't really understand the difference between `project` and just `sample`. In my own constraint samplers I have project just call sample :)

sachinchitta:
project and sample currently seem to do the same thing in all the constrained samplers but the original intent was, I think, for project to be used to "project" samples onto a constraint manifold.

regarding the solution here, a goal sampler needs to start from different random IK seeds to be able to successfully sample and do IK. The current implementation caused it to be using previous IK solutions every time, in which case it would always find the same (or really close) solution every time. This would not happen if the callback for the IK call was in there. 

Having said that, the use_state_as_seed does seem too IK specific to be in the base class for constraint samplers. We'll try and figure out a different way to do this. 

this is a fix though for a behavior we have been seen crop up in cramped spaces.

arjungm:
I agree that it is too IK specific. I came up with an alternate fix that does not require changing anything in moveit_core

Here: https://github.com/ros-planning/moveit_planners/pull/48

davetcoleman:
+1

Issue #181, opened 2014-07-16
Failing devel job breaks other prereleases


dirk-thomas:
Your devel job (http://jenkins.ros.org/job/devel-indigo-moveit_core/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/2/console) is failing due to wrong / incomplete CMake.

This also effects other package since it breaks prereleases as well, e.g.: http://jenkins.ros.org/job/prerelease-indigo-orocos_kinematics_dynamics/2/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=prerelease/console

Please fix the problems and rerelease the packages.

davetcoleman:
Looks like the package `angles` needs to be added to `find_package(catkin REQUIRED COMPONENTS)`? Currently its only `find_package(angles)` [here](https://github.com/ros-planning/moveit_core/blob/hydro-devel/CMakeLists.txt#L31)

Problem:

> /tmp/test_repositories/src_repository/moveit_core/constraint_samplers/test/pr2_arm_kinematics_plugin.h:45:27: fatal error: angles/angles.h: No such file or directory
>  #include <angles/angles.h>

k-okada:
I'm not sure, but you need to add agnles to `build_depend` if other build_depend package did not depends on angles
https://github.com/ros-planning/moveit_core/blob/hydro-devel/package.xml

;; from the next room...

davetcoleman:
I think this was fixed, above

dirk-thomas:
The referenced patch in #190 looks pretty wrong to me.

davetcoleman:
New fix attempted.

dirk-thomas:
This is still not fixed. The problem describes indigo but the correct patch was applied to hydro only (I did not even check if it was broken on hydro). But indigo still only contains the incomplete PR.

mikeferguson:
@dirk-thomas the devel job is now running but unstable, is that enough for this ticket?

(there are individual tickets already covering the tests that are (still) failing)

dirk-thomas:
Yes, this was all about the compile-from-source failing other downstream packages. The maintainers can handle the test failures however they want. Thanks.

PR #190, opened 2014-07-31
adding angles dependency

sachinchitta:


davetcoleman:
I assume this is for https://github.com/ros-planning/moveit_core/issues/181

dirk-thomas:
You can not `find_package(angles)` in the not-testing part of the CMakeLists file when you don't depend on the package. This will not work.

davetcoleman:
Good point. What is the correct way to include the package in CMakeLists for testing then?

dirk-thomas:
You `<test_depend>` on the package, call `find_package(NAME)` within you `CATKIN_TESTING_ENABLED` block and also in there use include dirs, libraries etc.

PR #191, opened 2014-08-01
Update package.xml

k-okada:
fix test_depend -> build_depend
fix for #190 ?


davetcoleman:
Oh I didn't see this until now. I think angles is only needed for testing, see my fix. https://github.com/ros-planning/moveit_core/pull/192

k-okada:
#192 is better

PR #192, opened 2014-08-01
Moved find_package angles to within CATKIN_ENABLE_TESTING

davetcoleman:
Fix for https://github.com/ros-planning/moveit_core/issues/181 on top of previous fix https://github.com/ros-planning/moveit_core/pull/190

Moved find_package of angles to sub-folder that uses it for constraint_samplers testing, inside of CATKIN_ENABLE_TESTING

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,6 @@ COMPONENTS
 
 find_package(Eigen REQUIRED)
 find_package(octomap REQUIRED)
-find_package(angles)
 
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(LIBFCL REQUIRED fcl)
--- a/constraint_samplers/CMakeLists.txt
+++ b/constraint_samplers/CMakeLists.txt
@@ -26,6 +26,8 @@ if(CATKIN_ENABLE_TESTING)
   find_package(orocos_kdl)
   find_package(angles)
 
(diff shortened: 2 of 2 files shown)


isucan:
+1

dirk-thomas:
The patch is not yet using the variables (include doors, etc) of angles.

davetcoleman:
>  (include doors, etc) of angles.

Doors??

dirk-thomas:
Include dirs.

davetcoleman:
I'm sorry, I don't have much experience with using find_package outside of catkin. Do you mean I should add the libraries something like this?

```
  target_link_libraries(test_constraint_samplers
    ${MOVEIT_LIB_NAME}
    ${catkin_LIBRARIES}
    ${orocos_kdl_LIBRARIES}
    ${angles_LIBRARIES}
  )
```

dirk-thomas:
Yes, and additionally before the great target:

```
include_directories(${angles_INCLUDE_DIRS})
```

`angles` might or might not have include directories and/or libraries. But that is a detail you shouldn't need to know about and just always use the variables.

davetcoleman:
Fixed.

sachinchitta:
This was supposed to be against indigo-devel - that was the failing version.

davetcoleman:
I just created a new PR for Indigo.

PR #194, opened 2014-08-04
Indigo devel angles fix

davetcoleman:
Fix for https://github.com/ros-planning/moveit_core/pull/192 where I merged to the wrong branch (Hydro instead of Indigo)

@dirk-thomas is this good?


dirk-thomas:
Lgtm

PR #198, opened 2014-09-04
Removed problematic assertion from default_constraint_samplers.cpp

hersh:
When doing a task where moveit is planning repeatedly, this assertion consistently fails for me after a few minutes of running.

Looking at the code, I see that the function it is in, callIK(), already returns a bool indicating success or failure, and callIK() is called from a sampling loop that keeps trying until success.  Therefore I don't see a problem with changing the "assert" to "return".


davetcoleman:
+1 makes a lot of sense to me

PR #199, opened 2014-09-05
Indigo devel Hydro merged

davetcoleman:
All these commits have already been merged into Hydro, I'm just moving them to Indigo since I switched to it today for my primary development


davetcoleman:
Reminder: ignore Travis CI on moveit_core because it is still running Precise and will always say:

> moveit_core: No definition of [liburdfdom-headers-dev] for OS version [precise]

PR #204, opened 2014-11-03
forward port #198 to indigo

mikeferguson:
As noted by @hersh on moveit-users: https://groups.google.com/forum/#!topic/moveit-users/KNlCGLXb46M

### DIFF
--- a/constraint_samplers/src/default_constraint_samplers.cpp
+++ b/constraint_samplers/src/default_constraint_samplers.cpp
@@ -571,9 +571,7 @@ bool constraint_samplers::IKConstraintSampler::callIK(const geometry_msgs::Pose 
       solution[ik_joint_bijection[i]] = ik_sol[i];
     state.setJointGroupPositions(jmg_, solution);
 
-    assert(validate(state));
-
-    return true;
+    return validate(state);
   }
   else
   {


mikeferguson:
@isucan We probably want to do a new indigo release -- this is a pretty major annoyance from what I've seen (moveit crashes pretty frequently)

isucan:
In the release we should be compiled in release mode,  so assert won't
bite,  but I will release soon

mikeferguson:
The assert was definitely biting in debs -- so we might want to also investigate the compile mode...

mikeferguson:
So, looks like the deb build includes "-DCMAKE_BUILD_TYPE=None" in the config (see http://jenkins.ros.org/job/ros-indigo-moveit-core_binarydeb_trusty_amd64/68/consoleFull) -- which I just tested locally, and that will cause the Release mode not to be set.

davetcoleman:
Do we know how to change that build type? I started building my local code in release mode for experiments and had incredible performance improvements.

mikeferguson:
It looks like we could convert:

```
if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
```

into:

```
if(NOT CMAKE_CONFIGURATION_TYPES AND (NOT CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE STREQUAL "None"))
```

mikeferguson:
If that seems amenable -- I can open PRs on all the relevant repos

davetcoleman:
I'm not sure where that if statement is coming from, but it seems like a good idea to me.

mikeferguson:
Ideally, we have something that allows you to set a global workspace DEBUG, hence the IF statement being there. I've tested this on several packages locally and it appears to be the right answer for the build farm setting CMAKE_BUILD_TYPE=None, although, it seems like maybe this should be fixed on the farm, since most packages use the first version of the IF statement -- thoughts @tfoote @dirk-thomas

tfoote:
Setting `CMAKE_BUILD_TYPE=None` seems odd to me. It reminds me of https://github.com/ros-infrastructure/bloom/pull/278 a lot @wjwwood

dirk-thomas:
The usage of `-DCMAKE_BUILD_TYPE=None` is very weird and seem to be the case for all packages on the farm: http://jenkins.ros.org/job/ros-indigo-roscpp_binarydeb_trusty_amd64/40/consoleFull

The current Debian package of `moveit_core` (0.6.11-0trusty-20141107-2351-+0000) has been released with bloom 0.5.12 and clearly has the packaging fix (https://github.com/ros-gbp/moveit_core-release/blob/debian/ros-indigo-moveit-core_0.6.11-0_trusty/debian/compat).

The library sizes are way to small to be a "Debug" or "None" build though which is the most confusing part.

Anyway I would highly recommend not to patch this in `moveit_core` for now. @wjwwood Can you please double check what bloom generates and how the packages are build on the farm? It looks like they are still not build with `CMAKE_BUILD_TYPE=Release`.

wjwwood:
We do not set it explicitly because deb_helper, more specifically `dpkg-buildflags`, is supposed to set all the compiler flags for you.  The default cmake flags include `-DCMAKE_BUILD_TYPE=None` which I assume is to make sure the `dpkg-buildflags` do not get overridden by CMake. This is the template:

https://github.com/ros-infrastructure/bloom/blob/8c4716aa39dd49cfb28204b5fc61b9784c914c57/bloom/generators/debian/templates/rules.em

We can add `-DCMAKE_BUILD_TYPE=Release` to the list of CMake arguments we already pass but I have no idea if that is a good idea, in principle it should be fine, but I'm always wary of changing the template for all packages.

On a related note, we recently found that `rviz`'s performance problems were due to no `-DNDEBUG` being passed in the debs, but the release mode made very little difference:

https://github.com/ros-visualization/rviz/issues/775#issuecomment-59185412

dirk-thomas:
Does it currently pass `NDEBUG`? The description above that the assert is happening for the Debian package indicates that it is not passed.

wjwwood:
No, I pass it specifically in the rules file for rviz (I patched it manually).

wjwwood:
https://github.com/ros-gbp/rviz-release/commit/66f3d90b6e2cf536192ea07eb8015edda3f3d0d7

(I had previously set the `CMAKE_BUILD_TYPE` to Release)

davetcoleman:
Should we add this `NDEBUG` rule to moveit_core's CMakeLists.txt or to the bloom release rules for moveit_core?

wjwwood:
@davetcoleman at this point it's up to the `moveit_core` developers, but we may decide to make it on by default for all packages depending on the outcome or ros-infrastructure/bloom#327.

PR #205, opened 2014-11-22
fix test: isEmpty() requires that robot_state be diff

### LINKED ISSUE #116: Test Broken: planning_scene LoadRestoreDiff
/home/schitta/code/moveit/src/moveit_core/planning_scene/test/test_planning_scene.cpp:82: Failure
Value of: planning_scene::PlanningScene::isEmpty(ps_msg)
  Actual: false
Expected: true
[  FAILED  ] PlanningScene.LoadRestoreDiff (747 ms)

mikeferguson:
Fixes #116. This was probably broken by https://github.com/ros-planning/moveit_core/commit/103a268a819f7f9e2c3adb929f5948ab62e44f7f in which isEmpty(RobotState) began to require is_diff to be true -- of course, the default value is false.

@isucan -- I chose to fix the test here -- but perhaps the definition of isEmpty() is logically wrong instead?

### DIFF
--- a/planning_scene/test/test_planning_scene.cpp
+++ b/planning_scene/test/test_planning_scene.cpp
@@ -90,6 +90,7 @@ TEST(PlanningScene, LoadRestoreDiff)
   world.addToObject("sphere", shapes::ShapeConstPtr(new shapes::Sphere(0.4)), id);
 
   moveit_msgs::PlanningScene ps_msg;
+  ps_msg.robot_state.is_diff = true;
   EXPECT_TRUE(planning_scene::PlanningScene::isEmpty(ps_msg));
   ps->getPlanningSceneMsg(ps_msg);
   ps->setPlanningSceneMsg(ps_msg);


mikeferguson:
@isucan did you happen to miss this one and #206?

isucan:
This one is a bit tough.  This was sone on purpose.  I forget the exact context this is important,  but the pick and place and publishing of scenes to rain will be affected. I worry this will introduce subtle bugs.  Why do we need this?

mikeferguson:
@isucan -- I'm only fixing the test here, the change to isEmpty was done a long time ago, I just brought up that maybe it should be reviewed in more detail.

isucan:
Yeah,  I  noticed after making the comment,  and then merged.  I think we
should leave isEmpty as it is for now.

mikeferguson:
Yeah, I concur on leaving it as-is -- I think we've finally found most (if not all) of the is_diff bugs at this point.

Issue #117, opened 2013-11-12
Test Broken: constraint_samplers


sachinchitta:
**\* glibc detected **\* /home/schitta/code/moveit/devel/lib/moveit_core/test_constraint_samplers: invalid fastbin entry (free): 0x0000000002b2df60 ***

mikeferguson:
fix merged

Issue #112, opened 2013-11-12
Test Broken: PositionConstraintsFixed


sachinchitta:
Error:   Link '' not found in model 'pr2'
         at line 1083 in /home/schitta/code/moveit/src/moveit_core/robot_model/src/robot_model.cpp
Warning: Position constraint link model  not found in kinematic model.  Constraint invalid.
         at line 278 in /home/schitta/code/moveit/src/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
Warning: No frame specified for position constraint on link 'l_wrist_roll_link'!
         at line 284 in /home/schitta/code/moveit/src/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
Warning: No frame specified for position constraint on link 'l_wrist_roll_link'!
         at line 284 in /home/schitta/code/moveit/src/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
Warning: No frame specified for position constraint on link 'l_wrist_roll_link'!
         at line 284 in /home/schitta/code/moveit/src/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
/home/schitta/code/moveit/src/moveit_core/kinematic_constraints/test/test_constraints.cpp:421: Failure
Value of: pc.decide(ks).satisfied
  Actual: false
Expected: true
/home/schitta/code/moveit/src/moveit_core/kinematic_constraints/test/test_constraints.cpp:436: Failure
Value of: pc.decide(ks).satisfied
  Actual: false
Expected: true
/home/schitta/code/moveit/src/moveit_core/kinematic_constraints/test/test_constraints.cpp:448: Failure
Value of: pc.decide(ks).satisfied
  Actual: false
Expected: true
[  FAILED  ] LoadPlanningModelsPr2.PositionConstraintsFixed (617 ms)

mikeferguson:
fix merged

PR #206, opened 2014-11-22
clean up, start to fix constraint sampler tests

mikeferguson:
This doesn't fix these tests, but it does cleanup hundreds (maybe even thousands) of "Returning dirty link transforms" messages, making the tests run quite a bit quicker, and making logs a bit easier to read.


mikeferguson:
Added another commit that fixes some of the constraint test failures. Commit eee44be3cb70dc1a8df209dc501f2a0d16080713 changed when valid() becomes true to be after the sampler is configured, so the test needed to be updated for that. Also updated the torso position to not be an uninitialized value (looks like it was never fully fixed from the robot_state update).

All constraint sampler tests except LoadPlanningModelsPr2.MixedJointAndIkSamplerManager now pass, which means #112, #117 will be fixed.

mikeferguson:
Once PRs through 209 are merged, LoadPlanningModelsPr2.MixedJointAndIkSamplerManager should be the only failing test (I still haven't actually figured out why it is failing or how to fix it)

Issue #111, opened 2013-11-12
Test Broken: AttachedBodyTester


sachinchitta:
/home/schitta/code/moveit/src/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp:320: Failure
Value of: res.collision
  Actual: false
Expected: true
[  FAILED  ] FclCollisionDetectionTester.AttachedBodyTester (759 ms)

mikeferguson:
fix merged

Issue #110, opened 2013-11-12
Test Broken: fcl FclCollisionDetectionTester.ContactPositions


sachinchitta:
/home/schitta/code/moveit/src/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp:257: Failure
Value of: res.collision
  Actual: false
Expected: true
[  FAILED  ] FclCollisionDetectionTester.ContactPositions (753 ms)

mikeferguson:
fix merged

Issue #109, opened 2013-11-12
Test Broken: FclCollisionDetectionTester.ContactReporting


sachinchitta:
/home/schitta/code/moveit/src/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp:217: Failure
Value of: 2
Expected: res.contacts.size()
Which is: 1
[  FAILED  ] FclCollisionDetectionTester.ContactReporting (779 ms)

mikeferguson:
fix merged

PR #209, opened 2014-11-25
fix fcl_collision_detection tests

### LINKED ISSUE #108: Test Broken: fcl FclCollisionDetectionTester.LinksInCollision
/home/schitta/code/moveit/src/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp:176: Failure
Value of: res3.collision
  Actual: false
Expected: true
[  FAILED  ] FclCollisionDetectionTester.LinksInCollision (755 ms)

mikeferguson:
This fixes #108 #109 #110 also stops segfault in test. Mostly a bunch of calls to robotstate.update(), but also fixed bad pose being passed in TestChangingShapeSize and missing link in gripper that was in collision.


mikeferguson:
Also fixes #111

PR #210, opened 2014-11-25
Fix truncated debug message

davetcoleman:
There is a bug that does not allow an entire robot model to be printed to debug logging because of a 1024 byte limit on the message:

https://github.com/ros/console_bridge/blob/master/src/console.cpp#L75

Instead you only get a truncated version of the model info. As a workaround I thought we could possibly use std::cout and surround it with NDEBUG as discussed in https://github.com/ros-infrastructure/bloom/issues/327#issuecomment-64295814

My fix here probably isn't right though, I welcome corrections and feedback. Thanks!

### DIFF
--- a/robot_model/src/robot_model.cpp
+++ b/robot_model/src/robot_model.cpp
@@ -113,9 +113,9 @@ void moveit::core::RobotModel::buildModel(const urdf::ModelInterface &urdf_model
     logDebug("... constructing joint group states"); 
     buildGroupStates(srdf_model);
 
-    std::stringstream ss;
-    printModelInfo(ss);
-    logDebug("%s", ss.str().c_str());
+    // For debugging entire model
+    if (false)
+      printModelInfo(std::cout);
   }
   else
     logWarn("No root link found");


isucan:
This is very sucky indeed. We can do this as a temp hack, but we should make that 1024 larger maybe? 4k?

davetcoleman:
That would require all calls to logX() on console_bridge allocate 4x more memory though, right? I've never seen any other debug output exceed that limit, so it seems it would be very wasteful. 

What would be even better is if console_birdge provided a string input or a stringstream.

But that is a lot of effort. What if I remove the compile flags and surround the `printModelInfo()` with `if (false)` so that it is never actually outputted but is still preserved for future debugging?

isucan:
the (false) thing sounds good to me

davetcoleman:
Updated and squashed

mikeferguson:
Sweet and now I can turn it on with:

```
 #define false true
```

davetcoleman:
giphy

PR #211, opened 2014-11-26
fix last test, remove final warnings from test compilation

mikeferguson:
The primary relevant line is the change from num_attempts=1 to num_attempts=100. It appears that all other IK/position/orientation tests use 100 except this one. This test does sometimes pass, but mostly fails unless num_attempts is at least 2. @isucan @davetcoleman any thoughts?

The other lines are simply left over cruft from previous refactors (mainly the robot_state one).

### DIFF
--- a/constraint_samplers/test/test_constraint_samplers.cpp
+++ b/constraint_samplers/test/test_constraint_samplers.cpp
@@ -649,8 +649,6 @@ TEST_F(LoadPlanningModelsPr2, PoseConstraintSamplerManager)
   ks_const.setToDefaultValues();
   ks_const.update();
 
-  robot_state::Transforms &tf = ps->getTransformsNonConst();
-
   kinematic_constraints::PositionConstraint pc(kmodel);
 
   moveit_msgs::PositionConstraint pcm;
@@ -901,11 +899,6 @@ TEST_F(LoadPlanningModelsPr2, MixedJointAndIkSamplerManager)
   ks_const.setToDefaultValues();
   ks_const.update();
 
-  robot_state::Transforms &tf = ps->getTransformsNonConst();
-
-  std::map<std::string, double> state_values;
-  //  ks.getStateValues(state_values);
-
   moveit_msgs::Constraints con;
   con.joint_constraints.resize(1);
 
@@ -959,12 +952,11 @@ TEST_F(LoadPlanningModelsPr2, MixedJointAndIkSamplerManager)
   ASSERT_TRUE(ucs);
 
   constraint_samplers::IKConstraintSampler* ikcs_test  = dynamic_cast<constraint_samplers::IKConstraintSampler*>(ucs->getSamplers()[1].get());
-
   ASSERT_TRUE(ikcs_test);
 
   for (int t = 0 ; t < 1; ++t)
   {
-    EXPECT_TRUE(s->sample(ks, ks_const, 1));
+    EXPECT_TRUE(s->sample(ks, ks_const, 100));
     EXPECT_TRUE(jc.decide(ks).satisfied);
     EXPECT_TRUE(ikcs_test->getPositionConstraint()->decide(ks).satisfied);
     EXPECT_TRUE(ikcs_test->getOrientationConstraint()->decide(ks).satisfied);
@@ -979,8 +971,6 @@ TEST_F(LoadPlanningModelsPr2, SubgroupJointConstraintsSamplerManager)
   robot_state::RobotState ks_const(kmodel);
   ks_const.setToDefaultValues();
   ks_const.update();
-
-  robot_state::Transforms &tf = ps->getTransformsNonConst();
 
   kinematic_constraints::JointConstraint jc1(kmodel);
   moveit_msgs::JointConstraint jcm1;


davetcoleman:
The refactoring is good, and the PR compiles, but I am unable to get the moveit_core tests to pass on my computer using catkin_tools... it seems to just run forever. I use the command:

```
catkin build moveit_core --verbose -i --no-deps --catkin-make-args run_tests
```

And get output similar to:

```
[moveit_core] DEBUG ros.moveit_core: Solution : 1                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 0: -0.188321                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 1: -0.019935                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 2: -0.418497                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 3: -1.278470                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 4: -1.798392                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 5: -1.249166                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 6: -3.016897                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core:                                                                                                                                                                                                           
[moveit_core] DEBUG ros.moveit_core:                                                                                                                                                                                                           
[moveit_core] DEBUG ros.moveit_core: 0.804528 3.750000 -0.650000 294 145                                                                                                                                                                       
[moveit_core]                                                                                                                                                                                                                                  
[moveit_core]                                                                                                                                                                                                                                  
[moveit_core] DEBUG ros.moveit_core: Solution : 1                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 0: 0.394459                                                                                                                                                                                               
[moveit_core] DEBUG ros.moveit_core: 1: -0.166220                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 2: 0.804528                                                                                                                                                                                               
[moveit_core] DEBUG ros.moveit_core: 3: -1.253684                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 4: 1.506210                                                                                                                                                                                               
[moveit_core] DEBUG ros.moveit_core: 5: -1.253707                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core: 6: -2.294185                                                                                                                                                                                              
[moveit_core] DEBUG ros.moveit_core:                                                                                                                                                                                                           
[moveit_core] DEBUG ros.moveit_core:                                                                                                                                                                                                           
[moveit_core] DEBUG ros.moveit_core: 3.216562 3.750000 -0.650000 53 386                                                                                                                                                                        
[moveit_core]                                                              
```

Any suggestions?

mikeferguson:
I don't use catkin_tools -- still on catkin_make. Have you tried running it without verbose? Prior to fixing all the robotState.update() issues, it took like 10 minutes for the tests to run with all the output (lots and lots of output in these tests, guessing that debug might be similar)

mikeferguson:
Ok, i just tried this with catkin_tools, and it runs fine on my machine -- but I'm not getting DEBUG messages either -- do you have some other setting set?

mikeferguson:
Green on the farm! http://jenkins.ros.org/job/devel-indigo-moveit_core/19/

davetcoleman:
I have all ROS nodes by default set to

```
log4j.logger.ros=DEBUG
```

And then I individually disable logs as they annoy me. It seems that setting really slowed down everything. I changed it to its default behavior (INFO) and it finished in less than minute.

Although I'm still seeing some stuff like

> ERROR ros.rosconsole_bridge.console_bridge: IK solver computes joint values for joint 'l_shoulder_pan_joint' but group 'base' does not contain such a joint. 
> 
> Warning: Orientation constraint for link 'r_wrist_roll_link' is probably incorrect: 0.000000, 0.000000, 0.000000, 0.000000. Assuming identity instead.
> 
> WARN ros.rosconsole_bridge.console_bridge: Returning dirty link transforms

What did you investigation find about those?  I pulled and built the latest from indigo-devel.

Thanks so much for all of this effort!

mikeferguson:
@davetcoleman there are a number of errors and warnings -- but they are all intentional as far as I can tell -- for instance, tests that ask for the wrong group and look to see that the function returns false. All of the asserts and expectations for all tests are being met.

And yes, there is a hugely absurd amount of debug data coming out during these tests since, for instance, the constraint sampler runs something on the order of 600 calls to sample(). Now that I added the proper calls to RobotState::update(), there is only about 5% of what used to be there, but it is still a lot.

mikeferguson:
Oh, I should also say, I never did find what was creating that last set of "returning dirty link transforms" in UnionConstraintSampler test. It doesn't appear to be causing the test any problems, but I don't see any other updates to the RobotState, so I can't figure out where update() is missing.

davetcoleman:
I fixed the verbose debug output problem I was having with tests:
https://github.com/ros-planning/moveit_core/pull/218

PR #214, opened 2014-12-02
moveit_core components of collision plugins

mikeferguson:
This is the moveit_core aspects of https://github.com/ros-planning/moveit_ros/pull/536. I put this in collision_detection rather than creating yet another folder, since it depends on many headers in collision_detection anyways.


davetcoleman:
+1

PR #216, opened 2014-12-11
Allow KinematicBase to have RobotModel passed in to save loading time

davetcoleman:
This allows some IK solvers, including the default KDL solver in MoveIt!, to load much faster by passing in a pre-loaded robot model instead of loading and parsing one from the parameter server each time an IK solver is initialized. This is important because multiple IK solvers are loaded, typically one for each planning group and again for each MoveIt! node. This does not currently improve loading time of IKFast solvers, however, because those robot models must be converted to Collada format and parsed by openrave. 

Changes:
- Passes a pointer to a RobotModel, or a Null pointer for backwards compatibility. 
- Deprecated initialize() function that does not use RobotModel
- Added `getParentModelPtr()` to joint model group
- Added `boost::enable_shared_from_this` to allow the shared pointer to be recieved by objects that only have its raw pointer

Additionally, KinematicsBase is now forward declaring RobotState and RobotModel, so @isucan suggested:

> To resolve cyclic dependency between libs, we should probably put everything in moveit_core into one lib.

I have not done this yet, but am thinking I could create a new library `moveit_core` that includes `moveit_robot_model`, `moveit_robot_state`, and `moveit_kinematics_base`. Thoughts? It still currently compiles without this last change though.

### DIFF
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -336,7 +336,7 @@ public:
                           const moveit::core::RobotModel* robot_model)
   {
     // For IK solvers that do not support passing in a robot_model pointer
-    return initialize(robot_description, group_name, base_frame, tip_frame, search_discretization, NULL);
+    return initialize(robot_description, group_name, base_frame, tip_frame, search_discretization);
   }
 
   /**
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -308,36 +308,14 @@ public:
    * @param tip_frame The tip of the chain
    * @param search_discretization The discretization of the search when the solver steps through the redundancy
    * @return True if initialization was successful, false otherwise
+   *
+   * Note: this is intended to be removed in future ROS versions in favor of robot_model being passed in
    */
-  MOVEIT_DEPRECATED
   virtual bool initialize(const std::string& robot_description,
                           const std::string& group_name,
                           const std::string& base_frame,
                           const std::string& tip_frame,
                           double search_discretization) = 0;
-
-  /**
-   * @brief  Initialization function for the kinematics, for use with kinematic chain IK solvers
-   * @param robot_description This parameter can be used as an identifier for the robot kinematics it is computed for;
-   * For example, the name of the ROS parameter that contains the robot description;
-   * @param group_name The group for which this solver is being configured
-   * @param base_frame The base frame in which all input poses are expected.
-   * This may (or may not) be the root frame of the chain that the solver operates on
-   * @param tip_frame The tip of the chain
-   * @param search_discretization The discretization of the search when the solver steps through the redundancy
-   * @param robot_model - intended to replace reliance on robot_description, this allows the URDF to be loaded much quicker
-   * @return True if initialization was successful, false otherwise
-   */
-  virtual bool initialize(const std::string& robot_description,
-                          const std::string& group_name,
-                          const std::string& base_frame,
-                          const std::string& tip_frame,
-                          double search_discretization,
-                          const moveit::core::RobotModel* robot_model)
-  {
-    // For IK solvers that do not support passing in a robot_model pointer
-    return initialize(robot_description, group_name, base_frame, tip_frame, search_discretization);
-  }
 
   /**
    * @brief  Initialization function for the kinematics, for use with non-chain IK solvers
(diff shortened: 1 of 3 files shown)


kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -313,6 +317,29 @@ class KinematicsBase
...
+   * @param search_discretization The discretization of the search when the solver steps through the redundancy
+   * @param robot_model - intended to replace reliance on robot_description, this allows the URDF to be loaded much quicker
+   * @return True if initialization was successful, false otherwise
+   */
+  virtual bool initialize(const std::string& robot_description,
+                          const std::string& group_name,
+                          const std::string& base_frame,
+                          const std::string& tip_frame,
+                          double search_discretization,
+                          const moveit::core::RobotModel* robot_model)
```
mikeferguson: should this be a smart pointer rather than a raw one?

davetcoleman: Kinematic plugins are loaded within the RobotModel constructor, so it is impossible to use a RobotModel smart pointer. I instead have to use `enable_shared_from_this` to get the smart pointer later. This feature was very tricky to implement.

mikeferguson: Ah that does make sense...

mikeferguson:
Does this break compilation or use of existing IKFast plugins? If not, +1 from me.

davetcoleman:
It does not break IKFast plugins, at least at compile time. I should probably do some more runtime testing

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -313,6 +317,29 @@ class KinematicsBase
...
+   * For example, the name of the ROS parameter that contains the robot description;
+   * @param group_name The group for which this solver is being configured
+   * @param base_frame The base frame in which all input poses are expected.
+   * This may (or may not) be the root frame of the chain that the solver operates on
+   * @param tip_frame The tip of the chain
+   * @param search_discretization The discretization of the search when the solver steps through the redundancy
+   * @param robot_model - intended to replace reliance on robot_description, this allows the URDF to be loaded much quicker
+   * @return True if initialization was successful, false otherwise
+   */
+  virtual bool initialize(const std::string& robot_description,
```
sachinchitta, line 333: I don't like this new function - if we don't need robot description (since we have robot model), we should not need to pass it in. I would redesign this to just use a RobotModel ptr (with group name and search discretization).

sachinchitta:
I don't use IKFast plugins as much but are there tests for those plugins? I would prefer this PR to be accompanied by tests implemented for those plugins first. Also, make sure that the tests that are already there for the PR2 pass (which you might have checked already). Changes in this particular component of MoveIt! affect everything else so its important to be careful.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman, line 333: See comments for moveit_ros PR

davetcoleman:
I do not believe there are tests for IKFast plugins, and I am not very motivated to create them since I also do not currently use IKFast (though I do help maintain the package). 

I haven't run PR2 tests yet, will do.

davetcoleman:
I ran the PR2 tests and they do pass.

davetcoleman:
One issue with this PR I currently have - when compiling I now always get a deprecated warning because I have the new function by default calling the old function (for compatibility):

> /home/dave/ros/ws_moveit/src/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h:339:98: warning: virtual bool kinematics::KinematicsBase::initialize(const string&, const string&, const string&, const string&, double) is deprecated (declared at /ho\
> me/dave/ros/ws_moveit/src/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h:313) [-Wdeprecated-declarations]
>      return initialize(robot_description, group_name, base_frame, tip_frame, search_discretization);

Is there a way to suppress the warning for that one instance?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
sachinchitta, line 333: another way to do this could be to have the new initialize function be simply: initialize(robot_model, search_discretization, group_name) - make it return false by default, check on that in kinematics_plugin loader and try the other (older) initialize function then. This way things will stay cleaner and separated and we won't break anyone using the old initialize function (yet). For ROS-J, I would honestly prefer writing a new plugin with an attempt to keep the base class as it is (as far as possible) - bandaging an old plugin can get painful after a while.

sachinchitta:
I have tried this PR and it broke MoveIt! in ways I could not understand - all the planning plugins disappeared. I agree that we need a way to initialize MoveIt! classes using a robot model passed in but this PR does not do it the right way.

davetcoleman:
@sachinchitta I do not understand why you closed this PR when you agree MoveIt! needs this functionality. Even though this PR did not get addressed for an entire year doesn't mean its bad.
- Could you explain more clearly what errors you experienced and maybe include some command line output?
- I'd also be curious to know what you think the right way is to do this PR? 

When I first created this PR I put in a lot of effort to make this backwards compatible, I tested it with several different IK solvers, and I ran the entire MoveIt! test suite. I've been using this feature for a year now in my own MoveIt fork and have had no issues.

I did just merge it with the latest jade and the basic MoveIt demo.launch is not working, likely because the recent KinematicBase multi-solution changes. I could address this. But if I fix it will there be support in getting it merged in or will you close it in another year?

sachinchitta:
I closed it because of the reasons mentioned when I tested it. These are the basic interaction tests I always do before merging any PR (not exhaustive):

Startup a demo.launch for a robot that I have (e.g. moveit_examples here has a Fanuc: https://github.com/sachinchitta/moveit_examples)
- move the robot around and make sure that works
- choose a start and goal position, change the planner to RRTConnect and create and visualize a plan

In this PR's case, all the planners disappeared. I did spend a bunch of time trying to figure it out - note that I have pulled in other PRs and updated them myself where I thought the functionality was useful and important - e.g. https://github.com/ros-planning/moveit_ros/pull/625.

I think this functionality is important and very significant but it feels ad-hoc and  weird to have RobotModel use kinematics base as a building block and vice-versa. I would prefer it if we rethink the relationship between the two in a more fundamental way - I think robot model should offer some kinematics solvers by default internally and expose a plugin model for other types of solvers.

PR #218, opened 2014-12-12
Namespaced pr2_arm_kinematics_plugin debug output

davetcoleman:
When running tests the pr2 kinematics plugin was very slow because of all the debug output being generated. This change allows me to diable it manually (by default on my machine I have all debug output _enabled_)

Does not change any functionality

Side note: why is ROS_DEBUG being used in moveit_core anyway?


mikeferguson:
+1, debug outputs should pretty much always be named IMO.

davetcoleman:
^^ agree very much

PR #219, opened 2014-12-13
Planning Scene: Ability to offset geometry loaded from stream

davetcoleman:
When loading a planning scene from file, this allows a pose to be provided to place that scene in a different location.


davetcoleman:
Thanks!

Issue #215, opened 2014-12-06
Storing distance field for collision detection


mikeferguson:
A number of planning algorithms (that are not yet or no longer in MoveIt) require a distance field. In the old SBPL interface, this was handled by having a different collision checker, which SBPL was hardcoded to use (or fallback and not use the distance field but have terrible performance). I'd like to avoid this in the future, which means having a standard storage spot for a distance field. So, where should we store this? I see three possible places:
- planning_scene::PlanningScene - We could hold a boost::shared_ptr to a distance_field in planning scene, making it always available to a number of things. The downside is that the planning_scene is already pretty cluttered with a lot of stuff.
- collision_detection::World -- collision geometry is stored here for collisions, which also includes the octomap. This class also already has all the callbacks coming in for change in objects. However, the distance_field is an alternate representation of the geometry that is already stored in World, so maybe this is not the best place to put it, but this location is what I'm leaning towards.
- collision_detection::CollisionWorld -- we could also store here, since the distance_field is an alternate representation of the data in World.

With any of these approaches, when the distance field isn't used, the pointer could be NULL (as is the case with the WorldDiff in PlanningScene) avoiding the cost of updating. With the new collision_plugin structure, we can easily setup the distance_field to be non-null when the collision checker wants a distance field.

A final possibility which avoids changing any of the above classes, is to create a new collision_detection::CollisionWorldWithDistanceField class that is derived from CollisionWorld and from which all distance-field based collision detectors are derived from, and then planning algorithms can try a dynamic_cast to this class and retrieve the field. I think the runtime penalty is not too high, as the only time the cast would occur is during setup for the motion planning request.

@isucan Any thoughts on this?

isucan:
@mikeferguson , this would be a great addition to MoveIt. I think the distance field should be in the PlanningScene. That is the object intended to contain all the necessary information for a planner to perform its computation. The DistanceField is a different concept than a collision checker, so I think it should be available directly. Of course, a collision detection wrapper could (should) be  implemented as well, for ease of use in planner. Does this make sense?

mikeferguson:
It does make sense to make it a part of the PlanningScene. I actually have spent a bit of time today looking more into the API, and I think that means we still have two options since the PlanningScene actually stores the geometric data about objects in a collision_detection::World instance, so:
[...]

PR #221, opened 2014-12-17
update distance_field API to no longer use geometry_msgs

mikeferguson:
This is a stepping stone on the way to #215. The distance_field API currently uses geometry_msgs::Pose, which it internally converts to an Eigen::Affine3d. collision_detection::World stores the poses as Eigen::Affine3d already, and then passes them through the notification system as such. Thus, we convert eigen -> geometry_msgs -> eigen every time an object is updated/moved/etc.

I can't find much code out there depending on the DistanceField API, but I marked as deprecated with an intent to remove these calls in Jade. I've already updated the tests to use the new API, and will also be updating the SBPL wrapper to use the new API before I get to opening a PR on that.

### DIFF
--- a/distance_field/include/moveit/distance_field/distance_field.h
+++ b/distance_field/include/moveit/distance_field/distance_field.h
@@ -37,6 +37,7 @@
 #ifndef MOVEIT_DISTANCE_FIELD_DISTANCE_FIELD_
 #define MOVEIT_DISTANCE_FIELD_DISTANCE_FIELD_
 
+#include <moveit/macros/deprecation.h>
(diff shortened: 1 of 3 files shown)


isucan:
Mike, this is great! I would encourage you to remove the deprecated API in indigo as well, if it is not used. Maybe send an e-mail to the mailing list asking if anyone uses the old API?

mikeferguson:
Ioan, I will drop an email to the mailing list and possibly remove deprecated API. Shortly after opening this PR, I also realized I need to break out a few other functions in this class in order to properly handle some object add/remove/move stuff (by getting and caching the previous points associated with an object).

PR #224, opened 2014-12-22
Add time factor to iterative_time_parametrization

skohlbr:
See https://github.com/ros-planning/moveit_ros/pull/547

### DIFF
--- a/trajectory_processing/src/iterative_time_parameterization.cpp
+++ b/trajectory_processing/src/iterative_time_parameterization.cpp
@@ -107,16 +107,21 @@ void IterativeParabolicTimeParameterization::applyVelocityConstraints(robot_traj
   const std::vector<int> &idx = group->getVariableIndexList();
   const robot_model::RobotModel &rmodel = group->getParentModel();
   const int num_points = rob_trajectory.getWayPointCount();
+  
+  double velocity_scaling_factor = 1.0;
+
+  if (max_velocity_scaling_factor > 0.0 && max_velocity_scaling_factor <= 1.0)  
+    velocity_scaling_factor = max_velocity_scaling_factor;
+  else
+    if (max_velocity_scaling_factor == 0.0)
+      logDebug("A max_velocity_scaling_factor of 0.0 was specified, defaulting to %f instead.", velocity_scaling_factor);
+    else
+      logWarn("Invalid max_velocity_scaling_factor %f specified, defaulting to %f instead.", max_velocity_scaling_factor, velocity_scaling_factor);
 
   for (int i = 0 ; i < num_points-1 ; ++i)
   {
     const robot_state::RobotStatePtr &curr_waypoint = rob_trajectory.getWayPointPtr(i);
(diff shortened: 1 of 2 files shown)


mikeferguson:
This PR seems fine if the others are good. Needs to be merged first.

sachinchitta:
Can you call this something like: max_velocity_scaling_factor since it only changes the max velocity and does not affect the accelerations? "time_factor" is too general to describe what this does.

isucan:
looks good to me

trajectory_processing/src/iterative_time_parameterization.cpp
```diff
@@ -111,13 +112,18 @@ void IterativeParabolicTimeParameterization::applyVelocityConstraints(robot_traj
   {
     const robot_state::RobotStatePtr &curr_waypoint = rob_trajectory.getWayPointPtr(i);
     const robot_state::RobotStatePtr &next_waypoint = rob_trajectory.getWayPointPtr(i+1);
-    
+
+    double velocity_scaling_factor = 1.0;
+
+    if (max_velocity_scaling_factor > 0.0 && max_velocity_scaling_factor <= 1.0)
+      velocity_scaling_factor = max_velocity_scaling_factor;
```
davetcoleman: This user error checking silently fails, I think it should produce a warning using logWarn. Or maybe we could remove this all together and just use an assert to improve performance.

skohlbr: Arg, I just noticed this is inside a loop, which of course is unnecessary. Will move it to top of that loop.

Using an assert only is dangerous, as funny things will happen when a value of 0.0 gets used (I did that accidentally, and the trajectory is complete garbage). Allowing 0.0 to automatically get mapped to 1.0 silently makes behavior completely transparent for users that update, but do not update the MotionPlanRequests they use (value will default to 0.0, get mapped to 1.0 and everything works as before). Adding a logwarn would mean that suddenly everybody would get warnings about not setting the max_velocity_scaling_factor in their MotionPlanRequest to a valid value. If this is desired is a matter of taste/policy.

mikeferguson: I concur with @skohlbr -- no assert, as 0.0 should be properly handled (we really don't want another instance of the "is_diff" nightmare where we are constantly having issues with default values of ROS messages). I would suggest that we only warn for values < 0 and greater than 1 if we want a warning.

skohlbr: Updated, see https://github.com/team-vigir/moveit_core/commit/58d1660e8bb540042e47a4168ad9221d5940f392. Now sets to 1.0 when 0.0 is incoming with logDebug output and provides logWarn output if specified factor is outside [0.0, 1.0](also using 1.0 default instead).

davetcoleman: Awesome, great work!

mikeferguson:
Whoops, I thought this failed because the moveit_msgs PR had not been merged... however, it appears that the farm just hiccup'd -- going to kick it to run again.

mikeferguson:
Ok, this builds fine...

PR #227, opened 2015-01-20
add ptr/const ptr types for distance field

mikeferguson:
Related to #215


isucan:
For consistency, you could also use MOVEIT_CLASS_FORWARD,

PR #231, opened 2015-02-25
KinematicsBase  changes

jrgnicho:
This PR follows from the discussion [here](https://groups.google.com/forum/#!searchin/moveit-users/Proposed$20Change$20to$20IK$20plugin$20definition/moveit-users/IGkUhU2CGu0/nHIMJ5TTOFIJ) in regards to extending KinematicsBase class in order to support finding/returning multiple IK solutions for a single tip pose. 
In summary, the important changes introduced here are as follows:
- Added a getMultipleIk(..) non pure virtual method which allows returning multiple joint solutions.  The default implementation here just makes a call to the getPositionIK method.  Implementations of this base class shall return several solutions when available.
- Created a "KinematicsSearch" enumeration containing flags for choosing how to discretize the redundant joints when finding the ik solutions for a tip pose.
- Created the  "KinematicsErrors" enumeration that provides supplementary error codes indicating the type of failure during a kinematics query.
- Added the setJointDiscretization() method that sets the discretization for each redundant joint.

### DIFF
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -56,9 +56,13 @@ class RobotState;
 namespace kinematics
 {
 
-namespace KinematicsSolution
+/*
+ * @enum KinematicSearches
+ * @brief Flags for choosing the type discretization applied on the redundancy during a multiple ik solution search
+ */
+namespace KinematicSearches
 {
-  enum KinematicsSolutionSearch
+  enum KinematicSearch
   {
     ONE = 1,
     ALL_DISCRETIZED ,
@@ -67,7 +71,23 @@ namespace KinematicsSolution
     ALL
   };
 }
-typedef KinematicsSolution::KinematicsSolutionSearch KinematicsSolutionSearch;
+typedef KinematicSearches::KinematicSearch KinematicSearch;
+
+/*
+ * @enum KinematicErrors
+ * @brief Kinematic error codes that occur in a multiple ik solution search
+ */
+namespace KinematicErrors
+{
+  enum KinematicError
+  {
+    OK = 1,
+    UNSUPORTED_SEARCH_REQUESTED,
+    DISCRETIZATION_NOT_INITIALIZED,
+    NO_SOLUTION
+  };
+}
+typedef KinematicErrors::KinematicError KinematicError;
 
 /**
  * @struct KinematicsQueryOptions
@@ -77,12 +97,24 @@ struct KinematicsQueryOptions
 {
   KinematicsQueryOptions() :
     lock_redundant_joints(false),
-    return_approximate_solution(false)
+    return_approximate_solution(false),
+  	solutions_search_code(KinematicSearches::ONE)
   {
   }
 
   bool lock_redundant_joints;
   bool return_approximate_solution;
+  KinematicSearch solutions_search_code;  // used when finding multiple solutions for the same pose
+};
+
+/*
+ * @struct KinematicsResult
+ * @brief Reports result details of an ik solution query
+ */
+struct KinematicsResult
+{
+	KinematicError kinematic_error;
+	double solution_percentage; // the percentage of solutions achieved over the total number of solutions searched.
 };
 
 
@@ -115,10 +147,20 @@ public:
                              moveit_msgs::MoveItErrorCodes &error_code,
                              const kinematics::KinematicsQueryOptions &options = kinematics::KinematicsQueryOptions()) const = 0;
 
+  /**
+   * @brief Given a desired pose of the end-effector, compute the set joint angles solutions that are able to reach it.  This
+   * is a default implementation that returns only one solution and so its result is equivalent to calling 'getPositionIK(...)'
+   * with a zero initialized seed.
+   * @param ik_pose the desired pose of the tip frame
+   * @param solutions a vector of vectors where each entry is a valid solution
+   * @param result a struct that reports the results of the query
+   * @param an option struct which contains the type of redundancy discretization used.  This default implementation only
+   * supports the KinmaticSearches::ONE discretization; requesting any other will result in failure.
+   * @return True if a valid set of solutions was found, false otherwise
+   */
   virtual bool getMultipleIK(const geometry_msgs::Pose &ik_pose,
                              std::vector< std::vector<double> >& solutions,
-                             moveit_msgs::MoveItErrorCodes &error_code,
-                             double& solution_percentage,
+                             KinematicsResult& result,
                              const kinematics::KinematicsQueryOptions &options) const;
 
   /**
@@ -471,11 +513,30 @@ public:
   }
 
   /**
+   * @brief Sets the discretization applied to each redundant joint
+   * @param discretization a map of joint indices and discretization value pairs.  The first pair element corresponds
+   * to the index of the joint to be discretized.
+   */
+  void setJointDiscretization(const std::map<int,double>& discretization)
+  {
+    redundant_joint_discretization_ = discretization;
+  }
+
+  /**
    * @brief  Get the value of the search discretization
    */
   double getSearchDiscretization() const
   {
     return search_discretization_;
+  }
+
+  /**
+   * @brief Returns the set of supported kinematics discretization search types.  This implementation only supports
+   * the KinematicSearches::ONE search.
+   */
+  std::vector<KinematicSearch> getSuportedSolutionSearches() const
+  {
+    return supported_searches_;
   }
 
   /** @brief For functions that require a timeout specified but one is not specified using arguments,
@@ -502,7 +563,9 @@ public:
                               // (if multiple tip frames provided, this variable will be unset)
     search_discretization_(DEFAULT_SEARCH_DISCRETIZATION),
     default_timeout_(DEFAULT_TIMEOUT)
-  {}
+  {
+    supported_searches_.push_back(KinematicSearches::ONE);
+  }
 
 protected:
 
@@ -515,6 +578,8 @@ protected:
   double search_discretization_;
   double default_timeout_;
   std::vector<unsigned int> redundant_joint_indices_;
+  std::map<int,double> redundant_joint_discretization_;
+  std::vector<KinematicSearch> supported_searches_;
 
 private:
 
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -56,6 +56,19 @@ class RobotState;
 namespace kinematics
 {
 
+namespace KinematicsSolution
+{
+  enum KinematicsSolutionSearch
+  {
+    ONE = 1,
+    ALL_DISCRETIZED ,
+    SOME_DISCRETIZED,
+    RANDOM_SAMPLE,
+    ALL
+  };
+}
+typedef KinematicsSolution::KinematicsSolutionSearch KinematicsSolutionSearch;
+
 /**
  * @struct KinematicsQueryOptions
  * @brief A set of options for the kinematics solver
@@ -101,6 +114,12 @@ public:
                              std::vector<double> &solution,
                              moveit_msgs::MoveItErrorCodes &error_code,
                              const kinematics::KinematicsQueryOptions &options = kinematics::KinematicsQueryOptions()) const = 0;
+
+  virtual bool getMultipleIK(const geometry_msgs::Pose &ik_pose,
+                             std::vector< std::vector<double> >& solutions,
+                             moveit_msgs::MoveItErrorCodes &error_code,
+                             double& solution_percentage,
+                             const kinematics::KinematicsQueryOptions &options) const;
 
   /**
    * @brief Given a desired pose of the end-effector, search for the joint angles required to reach it.
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -563,7 +563,7 @@ public:
     }
     else
     {
-      return 0.0f; // returned when there aren't any redundant joints
+      return 0.0; // returned when there aren't any redundant joints
     }
   }
 
@@ -571,7 +571,7 @@ public:
    * @brief Returns the set of supported kinematics discretization search types.  This implementation only supports
    * the DiscretizationMethods::ONE search.
    */
-  std::vector<DiscretizationMethod> getSuportedDiscretizationMethods() const
+  std::vector<DiscretizationMethod> getSupportedDiscretizationMethods() const
   {
     return supported_methods_;
   }
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -88,7 +88,7 @@ namespace KinematicErrors
     UNSUPORTED_DISCRETIZATION_REQUESTED,        /**< Discretization method isn't supported by this implementation */
     DISCRETIZATION_NOT_INITIALIZED,             /**< Discretization values for the redundancy has not been set. See
                                                      setSearchDiscretization(...) method*/
-    MULTIPLE_TIPS_NO_SUPPORTED,                 /**< Only single tip link support is allowed */
+    MULTIPLE_TIPS_NOT_SUPPORTED,                 /**< Only single tip link support is allowed */
     EMPTY_TIP_POSES,                            /**< Empty ik_poses array passed */
     NO_SOLUTION                                 /**< A valid joint solution that can reach this pose(s) could not be found */
 
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -90,6 +90,8 @@ namespace KinematicErrors
                                                      setSearchDiscretization(...) method*/
     MULTIPLE_TIPS_NOT_SUPPORTED,                 /**< Only single tip link support is allowed */
     EMPTY_TIP_POSES,                            /**< Empty ik_poses array passed */
+    IK_SEED_OUTSIDE_LIMITS,                                /**< Ik seed is out of bounds*/
+    SOLVER_NOT_ACTIVE,                          /**< Solver isn't active */
     NO_SOLUTION                                 /**< A valid joint solution that can reach this pose(s) could not be found */
 
   };
@@ -166,6 +168,7 @@ public:
    * 'getPositionIK(...)' with a zero initialized seed.
    *
    * @param ik_poses  The desired pose of each tip link
+   * @param ik_seed_state an initial guess solution for the inverse kinematics
    * @param solutions A vector of vectors where each entry is a valid joint solution
    * @param result A struct that reports the results of the query
    * @param options An option struct which contains the type of redundancy discretization used. This default
@@ -174,6 +177,7 @@ public:
    * @return True if a valid set of solutions was found, false otherwise.
    */
   virtual bool getPositionIK(const std::vector<geometry_msgs::Pose> &ik_poses,
+                             const std::vector<double> &ik_seed_state,
                              std::vector< std::vector<double> >& solutions,
                              KinematicsResult& result,
                              const kinematics::KinematicsQueryOptions &options) const;
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -165,11 +165,11 @@ public:
    * This is a default implementation that returns only one solution and so its result is equivalent to calling
    * 'getPositionIK(...)' with a zero initialized seed.
    *
-   * @param ik_pose The desired pose of the tip frame
+   * @param ik_poses  The desired pose of each tip link
    * @param solutions A vector of vectors where each entry is a valid joint solution
    * @param result A struct that reports the results of the query
    * @param options An option struct which contains the type of redundancy discretization used. This default
-   *                implementation only supports the KinmaticSearches::NO_DISCRETIZATION method; requesting any
+   *                implementation only supports the KinematicSearches::NO_DISCRETIZATION method; requesting any
    *                other will result in failure.
    * @return True if a valid set of solutions was found, false otherwise.
    */
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -57,34 +57,41 @@ namespace kinematics
 {
 
 /*
- * @enum KinematicSearches
- * @brief Flags for choosing the type discretization applied on the redundancy during a multiple ik solution search
+ * @enum DiscretizationMethods
+ *
+ * @brief Flags for choosing the type discretization method applied on the redundant joints during an ik query
  */
-namespace KinematicSearches
+namespace DiscretizationMethods
 {
-  enum KinematicSearch
+  enum DiscretizationMethod
   {
-    ONE = 1,
-    ALL_DISCRETIZED ,
-    SOME_DISCRETIZED,
-    RANDOM_SAMPLE,
-    ALL
+    NO_DISCRETIZATION = 1,      /**< The redundant joints will be fixed at their current value. */
+    ALL_DISCRETIZED ,           /**< All redundant joints will be discretized uniformly */
+    SOME_DISCRETIZED,           /**< Some redundant joints will be discretized uniformly. The unused redundant joints will be fixed at their
+                                     current value */
+    ALL_RANDOM_SAMPLED ,        /**< the discretization for each redundant joint will be randomly generated.*/
+    SOME_RANDOM_SAMPLED         /**< the discretization for some redundant joint will be randomly generated.
+                                     The unused redundant joints will be fixed at their current value. */
   };
 }
-typedef KinematicSearches::KinematicSearch KinematicSearch;
+typedef DiscretizationMethods::DiscretizationMethod DiscretizationMethod;
 
 /*
  * @enum KinematicErrors
- * @brief Kinematic error codes that occur in a multiple ik solution search
+ * @brief Kinematic error codes that occur in a ik query
  */
 namespace KinematicErrors
 {
   enum KinematicError
   {
-    OK = 1,
-    UNSUPORTED_SEARCH_REQUESTED,
-    DISCRETIZATION_NOT_INITIALIZED,
-    NO_SOLUTION
+    OK = 1,                                     /**< No errors*/
+    UNSUPORTED_DISCRETIZATION_REQUESTED,        /**< Discretization method isn't supported by this implementation */
+    DISCRETIZATION_NOT_INITIALIZED,             /**< Discretization values for the redundancy has not been set. See
+                                                     setSearchDiscretization(...) method*/
+    MULTIPLE_TIPS_NO_SUPPORTED,                 /**< Only single tip link support is allowed */
+    EMPTY_TIP_POSES,                            /**< Empty ik_poses array passed */
+    NO_SOLUTION                                 /**< A valid joint solution that can reach this pose(s) could not be found */
+
   };
 }
 typedef KinematicErrors::KinematicError KinematicError;
@@ -98,25 +105,30 @@ struct KinematicsQueryOptions
   KinematicsQueryOptions() :
     lock_redundant_joints(false),
     return_approximate_solution(false),
-  	solutions_search_code(KinematicSearches::ONE)
+    discretization_method(DiscretizationMethods::NO_DISCRETIZATION)
   {
   }
 
-  bool lock_redundant_joints;
-  bool return_approximate_solution;
-  KinematicSearch solutions_search_code;  // used when finding multiple solutions for the same pose
+  bool lock_redundant_joints;                   /**<  KinematicsQueryOptions#lock_redundant_joints. */
+  bool return_approximate_solution;             /**<  KinematicsQueryOptions#return_approximate_solution. */
+  DiscretizationMethod discretization_method;   /**< Enumeration value that indicates the method for discretizing the redundant. joints KinematicsQueryOptions#discretization_method. */
 };
 
 /*
  * @struct KinematicsResult
- * @brief Reports result details of an ik solution query
+ * @brief Reports result details of an ik query
+ *
+ * This struct is used as an output argument of the getPositionIK(...) method that returns multiple joint solutions.
+ * It contains the type of error that led to a failure or KinematicErrors::OK when a set of joint solutions is found.
+ * The solution percentage shall provide a ration of solutions found over solutions searched.
+ *
  */
 struct KinematicsResult
 {
-	KinematicError kinematic_error;
-	double solution_percentage; // the percentage of solutions achieved over the total number of solutions searched.
+	KinematicError kinematic_error;         /**< Error code that indicates the type of failure */
+	double solution_percentage;             /**< The percentage of solutions achieved over the total number
+	                                             of solutions explored. */
 };
-
 
 /**
  * @class KinematicsBase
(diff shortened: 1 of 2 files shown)


kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,39 @@ class RobotState;
...
 
+/*
+ * @enum KinematicSearches
+ * @brief Flags for choosing the type discretization applied on the redundancy during a multiple ik solution search
+ */
+namespace KinematicSearches
+{
+  enum KinematicSearch
+  {
+    ONE = 1,
```
shaun-edwards: I don't know that the meaning of these values is obvious.  Each enum value must be clearly defined so that plugin implementers know what they mean.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,39 @@ class RobotState;
...
+
+/*
+ * @enum KinematicErrors
+ * @brief Kinematic error codes that occur in a multiple ik solution search
+ */
+namespace KinematicErrors
+{
+  enum KinematicError
+  {
+    OK = 1,
```
shaun-edwards: Same here with defining the details of the failure enum (although these are more obvious than above).

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -64,12 +97,24 @@ struct KinematicsQueryOptions
 {
   KinematicsQueryOptions() :
     lock_redundant_joints(false),
-    return_approximate_solution(false)
+    return_approximate_solution(false),
+  	solutions_search_code(KinematicSearches::ONE)
```
shaun-edwards: spacing?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -64,12 +97,24 @@ struct KinematicsQueryOptions
...
+};
+
+/*
+ * @struct KinematicsResult
+ * @brief Reports result details of an ik solution query
+ */
+struct KinematicsResult
+{
+	KinematicError kinematic_error;
+	double solution_percentage; // the percentage of solutions achieved over the total number of solutions searched.
```
shaun-edwards: Should be a doxygen comment.  I don't think these members can be commented enough.  I would suggest that more comments with examples are needed.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -496,6 +578,8 @@ class KinematicsBase
   double search_discretization_;
   double default_timeout_;
   std::vector<unsigned int> redundant_joint_indices_;
+  std::map<int,double> redundant_joint_discretization_;
```
shaun-edwards, line 628: Is this really new?  How did iksolvers deal with >6DOF arms before?

kinematics_base/src/kinematics_base.cpp
```diff
@@ -122,3 +122,35 @@ const bool kinematics::KinematicsBase::supportsGroup(const moveit::core::JointMo
...
+                           const kinematics::KinematicsQueryOptions &options) const
+{
+  std::vector<double> solution;
+  std::vector<double> seed(getJointNames().size(),0.0f);
+  result.solution_percentage = 0.0f;
+
+  if(options.solutions_search_code != KinematicSearches::ONE)
+  {
+    result.kinematic_error = kinematics::KinematicErrors::UNSUPORTED_SEARCH_REQUESTED;
+    return false;
```
shaun-edwards, line 151: Should we log something here?  Not sure if logging from a plugin is good practice.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: +1

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: +1

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: +1

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: +1

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho, line 628: My guess is that the "search_discretization_" member was used for all of the joints whose indices were listed in the "redundant_joint_indices_" array.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,39 @@ class RobotState;
...
+ */
+namespace KinematicSearches
+{
+  enum KinematicSearch
+  {
+    ONE = 1,
+    ALL_DISCRETIZED ,
+    SOME_DISCRETIZED,
+    RANDOM_SAMPLE,
+    ALL
```
sachinchitta: This seems very specific to redundant joints. Its also not something most plugins may implement. Why not let this live in the plugins themselves - is there a case where you have to use one of these methods for some calls and then another one later on?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,39 @@ class RobotState;
...
+ * @brief Kinematic error codes that occur in a multiple ik solution search
+ */
+namespace KinematicErrors
+{
+  enum KinematicError
+  {
+    OK = 1,
+    UNSUPORTED_SEARCH_REQUESTED,
+    DISCRETIZATION_NOT_INITIALIZED,
+    NO_SOLUTION
```
sachinchitta: We used to have kinematics error codes earlier and then ditched them in favor of just MoveIt error/result flags. Again these codes seem very specific to the discretized search for redundant joints.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -103,6 +148,22 @@ class KinematicsBase
...
+   * @param result a struct that reports the results of the query
+   * @param an option struct which contains the type of redundancy discretization used.  This default implementation only
+   * supports the KinmaticSearches::ONE discretization; requesting any other will result in failure.
+   * @return True if a valid set of solutions was found, false otherwise
+   */
+  virtual bool getMultipleIK(const geometry_msgs::Pose &ik_pose,
+                             std::vector< std::vector<double> >& solutions,
+                             KinematicsResult& result,
+                             const kinematics::KinematicsQueryOptions &options) const;
+
```
sachinchitta, line 184: I can understand this function for 6 DOF robots - gives you elbow down, elbow up solutions, etc. What is this function expected to do for redundant robots - the notion of "all" solutions is ill-defined here since technically there are infinite solutions.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -452,6 +513,16 @@ class KinematicsBase
   }
 
   /**
+   * @brief Sets the discretization applied to each redundant joint
+   * @param discretization a map of joint indices and discretization value pairs.  The first pair element corresponds
+   * to the index of the joint to be discretized.
```
sachinchitta: What do the other pairs signify if only the first pair element is being used?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: While it's true that most plugins will hide the details of how the redundant joint are manipulated, there was a lot of interest in exposing this capability at the interface in the discussion which led to this PR.  
I could envision a case where in the interest of time one could try to get fewer solutions by discretizing fewer joints while in another case during path planning you'd like to explore larger solution sets in order to generate smoother transitions between points.  These larger sets would of course require more time to generate since more or all redundant joints would be discretized.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho, line 184: Agreed, for redundant arms the 'ALL' flag may not be applicable.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: I meant to say that in each pair, the key is the index and the value is the discretization for the joint that corresponds to that index.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
shaun-edwards: @sachinchitta, can you provide a link to the error flags (google can't seem to find them for me).  Also, can you suggest other error conditions that we should consider?

shaun-edwards:
Friendly ping.....I'd like to get this accepted as soon as possible.  We cannot integrate [Descartes](https://github.com/ros-industrial-consortium/descartes) into MoveIt without this change.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
sachinchitta: https://github.com/ros-planning/moveit_msgs/blob/indigo-devel/msg/MoveItErrorCodes.msg

Error codes are useful when you are going to do something in response to them or can modify your call in some way in response to the request, or for introspection. NO_IK_SOLUTION already exists in MoveIt!, so does success. There is nothing for discretization but I am reluctant to add even more codes that people have to deal with. Would prefer these to all live in one place.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
sachinchitta: Do you actually need to switch between these different flags to have Descartes working? We have done a planner similar to Descartes without needing these additional flags so that's why I am curious as to why you need them.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
sachinchitta, line 184: I would suggest that this function should have the following signature with a default that just returns the single solution. 

```
virtual bool getPositionIK(const geometry_msgs::Pose &ik_pose,
const std::vector<double> &ik_seed_state,
std::vector< std::vector<double> > &solution,
moveit_msgs::MoveItErrorCodes &error_code,
const kinematics::KinematicsQueryOptions &options =   kinematics::KinematicsQueryOptions()) const = 0;
```

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
sachinchitta: I have always used a single value for the search discretization. This parameter is really most useful when dealing with 7 DOF robots and when using an analytical solution - you will notice that the KDL kinematics solution does not use the search discretization. For greater than 7 DOF, you are probably better off using random seed positions anyways. Do you have a use case where you have more than one redundant joint whose range you need to search across?

sachinchitta:
Is there a minimal set of changes you can condense this down to for Descartes to work based on my feedback above?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
shaun-edwards: At a bare minimum, we need a method that returns _All_ IK solutions for a given pose.  For 6DOF robots, this is a fixed set of configurations (as you point out).  What this PR tries to address is how you control/configure the what _All_ means when there is redundancy.  This requirement to be able to configure the IK is driven form the conversation on the MoveIt list.  That being said, we already have a use case within Descartes where a rough search is performed, followed by a fine search if required, so there is a need.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
shaun-edwards: We have 7DOF robots attached to linear rails.  It's this exact use case that requires different discretization.  A linear rail might be 0-10m, and a revolute jiont -Pi - PI rads.  The difference in units and/or scale necessitates a different discretization.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
shaun-edwards: I think adding codes to the existing error codes is possible.  If we group all IK failures, whether they are caused by disretization or not, into a single error code, then the only possible error handling is too generic to be of much use.  From a user/developer standpoint, it would help to know if the failure is due to discretization or if the pose is in fact unreachable.

sachinchitta:
Can you point me to the Descartes branch with which these changes are integrated so I can have a look and see how they are being used?

shaun-edwards:
We have not integrated these specific changes.  Currently, Descartes wraps the ik plugin [here](https://github.com/ros-industrial-consortium/descartes/blob/hydro-devel/descartes_moveit/src/moveit_state_adapter.cpp#L152).  While this wrapper works for all plugins, it's assumed to be KDL.  When an IKFast solution is available, we directly link to the IKFast library.  See [here](https://github.com/ros-industrial/fanuc/blob/hydro-devel/fanuc_m16ib_moveit_plugins/m16ib20_kinematics/src/fanuc_m16ib20_manipulator_ikfast_moveit_plugin.cpp#L789) for example.  It already has a method that returns multiple joint solutions (at least in the case of 6DOF arms).  The IKFast functionality, which is hidden in the plugin, is what we would like to have access to.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman: +1 adding comments from the email discussion for reach enum value meaning

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -103,6 +148,22 @@ class KinematicsBase
...
+   * is a default implementation that returns only one solution and so its result is equivalent to calling 'getPositionIK(...)'
+   * with a zero initialized seed.
+   * @param ik_pose the desired pose of the tip frame
+   * @param solutions a vector of vectors where each entry is a valid solution
+   * @param result a struct that reports the results of the query
+   * @param an option struct which contains the type of redundancy discretization used.  This default implementation only
+   * supports the KinmaticSearches::ONE discretization; requesting any other will result in failure.
+   * @return True if a valid set of solutions was found, false otherwise
+   */
+  virtual bool getMultipleIK(const geometry_msgs::Pose &ik_pose,
```
davetcoleman: This should really accept multiple poses, which means all you need is the first parameter to simply be:

```
const std::vector<geometry_msgs::Pose> &ik_poses
```

Otherwise this is a digression of the work we've done to have the KinematicsBase API support humanoids. For example, see the member variables of this file:

```
  std::vector<std::string> tip_frames_;
  std::string tip_frame_; // DEPRECATED - this variable only still exists for backwards compatibility with
                          // previously generated custom ik solvers like IKFast
```

Your specific plugin can just throw an error if more than 1 pose is passed in. The ability to check for supporting multiple poses is in this function:

```
  virtual const bool supportsGroup(const moveit::core::JointModelGroup *jmg,
                                   std::string* error_text_out = NULL) const;
```

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
davetcoleman: +1 I also have a gantry-mounted robot that needs this

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
sachinchitta: I agree with Dave here about making sure the new support for Humanoids is not broken.

sachinchitta:
@shaun-edwards @jrgnicho - in the absence of a working integration with Descartes, I am reluctant to merge these changes. The main reason is not to have to break things again in the future when we find that we need more changes in the base class to make Descartes work - I know it looks like this should all work but in general we won't know until its all tested. I would prefer that you branch Descartes, integrate with the changes here, test and confirm so we know it all works. Also, if it looks like the main blocker is an IKFast style function that returns "all" solutions, that's much easier to integrate and merge quicker. I also see your point about requiring different discretization parameters for different types of joints - I think we should get that change in but also deprecate the existing discretization parameter so people don't get confused by all the options available. Also wanted to make sure that people understand that the discretized search functionality does not exist in the KDL solver right now. It exists in the PR2 solver and the individual IKFast solvers but only for a single redundant joint.

shaun-edwards:
@sachinchitta - That is fair.   Our rationale for submitting this PR before we integrate into Descartes is that there would be more interested parties in MoveIt and thus more feedback.  Since we control Descartes at this early stage, we can adapt to what the MoveIt community requires.  We will begin integration with Descartes under the assumption that this PR is more or less agreed to (keeping in mind the current feedback about humanoid support and discretization backwards compatibility).  Thanks for your efforts.

sachinchitta:
@shaun-edwards - thanks for the contribution - I can help coordinate the changes on the MoveIt! side as you work through the integration. In general, writing tests helps in making sure current functionality stays working so any tests you can contribute to help do that would also be welcome.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: It may be rather difficult to support multiple tips through this method because it'd be unclear to deternime which solutions (more than 1 per pose) belong to each tip pose.

davetcoleman: Your IK solver is welcome to only support 1 tip / kinematic chain, but this functionality is needed by some.

I just clarified the documentation https://github.com/ros-planning/moveit_core/pull/232

jrgnicho: I was confusing this multiple tips case with multiple kinematic chains case, but for humanoid/bipedal robots multiple tips can be part of the same chain.  +1 for supporting multiple tips

jrgnicho:
A unit test for the changes to the KinematicsBase class in this PR has been created [here](https://github.com/jrgnicho/descartes_moveit_test)

davetcoleman:
Wow, nice tests!

shaun-edwards:
@ping....our project is coming to an end.  If we are to continue with this PR, we must do so in the next few weeks.

sachinchitta:
At this point, it makes more sense for this to be integrated into jade since putting this into indigo will kill all the current plugins. Is it ok with you if we do this just around jade - the additional updates that will be required:
(1) update KDL plugin and its tests
(2) update IKFast plugins and the MoveIt! IKFast interface and incorporate your tests
(3) update PR2 plugin and its tests
(4) update some calls from MoveIt! into the kinematics solvers

jrgnicho:
@sachinchitta I'll submit PRs to those repositories with the required changes.  I don't quite get item 4, could you elaborate some more on that? Thanks for your guidance.

sachinchitta:
Hmm, I might be mistaken on (4). Let's work through (1), (2), (3) and see if there's any changes needed inside MoveIt! as we do that.

jrgnicho:
Hello Sachin,
I apologize for the delay on producing the required unit tests, it's been a hectic month for me.  I have added the test for the PR2, ikfast and kdl [here](https://github.com/jrgnicho/kinematics_base_tests).  However, there were a few changes I had to make in the [robot_interaction::KinematicsOption](https://github.com/jrgnicho/moveit_ros/blob/kinematic_base_changes/robot_interaction/include/moveit/robot_interaction/kinematic_options.h#L64) struct and so I'll create a PR for those.  Also, I will make another PR for the related changes to the ikfast plugin.  Since there are going to  be several associated PRs, I was wondering if there's an efficient/preferred way of doing this?

jrgnicho:
I've completed writing the [unit test](https://github.com/jrgnicho/kinematics_base_tests) for the ikfast, kdl and pr2 plugins.  Along with that, I made modifications in the moveit_ikfast and moveit_ros repositories to accomodate the changes in **kinematics::KinematicsBase** and placed them in branches under forked repos [here](https://github.com/jrgnicho/moveit_ikfast.git) and [here](https://github.com/jrgnicho/moveit_ros.git).  Is there a way to link PRs across multiple repos as part of the same change?

sachinchitta:
AFAIK, there is no easy way to put PRs together, we'll just have to pull things in at one time. Also, we will have to aim this at jade and not indigo. I will look at your changes and add comments over the next few days.

jrgnicho:
Ok, then should I close this PR and create a new one for jade?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,46 @@ class RobotState;
...
+ */
+namespace KinematicErrors
+{
+  enum KinematicError
+  {
+    OK = 1,                                     /**< No errors*/
+    UNSUPORTED_DISCRETIZATION_REQUESTED,        /**< Discretization method isn't supported by this implementation */
+    DISCRETIZATION_NOT_INITIALIZED,             /**< Discretization values for the redundancy has not been set. See
+                                                     setSearchDiscretization(...) method*/
+    MULTIPLE_TIPS_NO_SUPPORTED,                 /**< Only single tip link support is allowed */
```
gavanderhoorn: Typo? `NO` -> `NOT`.

jrgnicho: @gavanderhoorn good catch.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -444,19 +521,59 @@ class KinematicsBase
...
+  double getSearchDiscretization(int joint_index = 0) const
   {
-    return search_discretization_;
+    if(redundant_joint_discretization_.count(joint_index) > 0)
+    {
+      return redundant_joint_discretization_.at(joint_index);
+    }
+    else
+    {
+      return 0.0f; // returned when there aren't any redundant joints
```
davetcoleman: Why do you add an `f` to the end of this number (specifying it as a float) when the function returns a double? Seems to me it should be removed.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -444,19 +521,59 @@ class KinematicsBase
...
+    }
+  }
+
+  /**
+   * @brief Returns the set of supported kinematics discretization search types.  This implementation only supports
+   * the DiscretizationMethods::ONE search.
+   */
+  std::vector<DiscretizationMethod> getSuportedDiscretizationMethods() const
```
davetcoleman: function name `Supported` should have two p's

kinematics_base/src/kinematics_base.cpp
```diff
@@ -122,3 +126,52 @@ const bool kinematics::KinematicsBase::supportsGroup(const moveit::core::JointMo
...
   return true;
 }
+
+bool kinematics::KinematicsBase::getPositionIK(const std::vector<geometry_msgs::Pose> &ik_poses,
+                           std::vector< std::vector<double> >& solutions,
+                           kinematics::KinematicsResult& result,
+                           const kinematics::KinematicsQueryOptions &options) const
+{
+  std::vector<double> solution;
+  std::vector<double> seed(getJointNames().size(),0.0f);
```
davetcoleman: i see the `f` here again

davetcoleman:
I really appreciate the work you are putting into this, we should get it merged asap since Indigo is about to be released

jrgnicho:
@davetcoleman , thanks for your review.  The f's are just a bad habit, I went ahead and removed them.

jrgnicho:
The PRs to [moveit_ikfast](https://github.com/ros-planning/moveit_ikfast/pull/50) and [moveit_ros](https://github.com/ros-planning/moveit_ros/pull/581) that are related to this change have been created.

shaun-edwards:
@sachinchitta, please create a `jade` branch, so we can reissue the PR.

sachinchitta:
Done.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -444,19 +525,59 @@ class KinematicsBase
...
+  }
+
+  /**
+   * @brief Sets individual discretization values for each redundant joint.
+   *
+   * Calling this method replaces previous discretization settings.
+   *
+   * @param discretization a map of joint indices and discretization value pairs.
+   */
+  void setSearchDiscretization(const std::map<int,double>& discretization)
```
sachinchitta, line 547: I feel like this lets users setup the wrong number of joints as redundant, e.g. setup 3 joints as redundant for a 7 DOF arm. We should think about ways to prevent that.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -444,19 +525,59 @@ class KinematicsBase
...
+      redundant_joint_discretization_.insert(*i);
+      redundant_joint_indices_.push_back(i->first);
+    }
   }
 
   /**
    * @brief  Get the value of the search discretization
    */
-  double getSearchDiscretization() const
+  double getSearchDiscretization(int joint_index = 0) const
```
sachinchitta, line 562: I would prefer just returning the stored map instead of a single value like this. Its cleaner.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,48 @@ class RobotState;
 namespace kinematics
 {
 
+/*
+ * @enum DiscretizationMethods
+ *
+ * @brief Flags for choosing the type discretization method applied on the redundant joints during an ik query
+ */
+namespace DiscretizationMethods
```
sachinchitta, line 64: why separate namespaces? If we do use them, they should match the ROS style for namespaces.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -56,6 +56,48 @@ class RobotState;
...
+                                     The unused redundant joints will be fixed at their current value. */
+  };
+}
+typedef DiscretizationMethods::DiscretizationMethod DiscretizationMethod;
+
+/*
+ * @enum KinematicErrors
+ * @brief Kinematic error codes that occur in a ik query
+ */
+namespace KinematicErrors
```
sachinchitta, line 83: Again, why use a separate namespace?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho, line 547: Is it always up to the KinematicsBase implementation to set the redundant joints? If that's the case then we can check the keys in the input discretization map instead resetting the joints from it.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho, line 64: As far as namespaces to encapsulate enumerations I followed what's described [here](http://wiki.ros.org/CppStyleGuide#Enumerations), should I follow a different convention?

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho, line 83: See above

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jrgnicho: I modified this method in such way so that it did't break the existing implementations.  A new one that returns a const map can be added but that feels redundant.

jrgnicho, line 562: This might not be such an easy change. Replacing this method with something that returns a map breaks some code in moveit_ros (see below)

```
src/moveit_ros/planning/planning_components_tools/src/kinematics_speed_and_validity_evaluator.cpp:72:75: note:   ‘const std::map<int, double>’ is not derived from ‘const std::basic_string<_CharT, _Traits, _Alloc>’
         ROS_INFO_STREAM("Search res: " << solver->getSearchDiscretization())
```

An extra method that returns a const reference to the map could be added but that feels slightly redundant.

shaun-edwards:
@jrgnicho, should this PR be closed since these changes appear to have been accepted in another PR (#248)

jrgnicho:
@shaun-edwards  I think these changes should still be merge into indigo-devel since the corresponding PR in moveit_ros was [merged](https://github.com/ros-planning/moveit_ros/pull/581) into that repo's indigo-devel branch.  Not doing so breaks the moveit_ros build.  @sachinchitta, any thoughts on this?

mikeferguson:
OK, so I'm back after being missing for a bit here with company launch and all -- but it seems we have a bit of a mess right now. moveit_ros will not build in indigo-devel -- and we are already having issues syncing indigo and jade. Honestly, indigo-devel (as indigo is an LTS) will outlast jade-devel. As we are below 1.0, we don't have to make any assurances about APIs staying the same.

90% of everything in Jade is being released from an indigo-devel branch, unless there is a compelling break they have to do  (for instance, navigation is > 1.0, and we changed some plugin APIs).

Given all of this, and the fact we already broke the  build, we should really merge this, backport the one or two changes that made it into jade-devel alone, kill the jade-devel branches, and get a new set of releases into indigo and jade.

I have set aside some time over the next week, I can make my way through the entire moveit stack and make sure we get this properly fixed up and released -- any objections? @shaun-edwards @sachinchitta @davetcoleman @isucan

mikeferguson:
Also, meant to say, I'm sure a few people will have a bit of a headache fixing their IKFast plugins -- however, I think even more people will have headaches if we basically abandon the indigo-devel branch and put new changes only in Jade (as I expect most users will stay with Indigo, especially when working with real robots)

sachinchitta:
I think we should stick with the original plan and merge only into jade-devel. This merge into indigo-devel was a mistake, we need to pull it out.

mikeferguson:
So, what is the reason not to merge this? Looking closely at it -- IKFast plugins will need to be recompiled, but they should still compile, right? The kinematics base class provides a default implementation of each added function, so it should still compile, right? They may not work with a new planner that calls the particular getPositionIK, but that doesn't mean that backwards compatibility is not maintained.

If you have a catkin workspace, you should always rebuild after updating debs since ABI is never guaranteed, and if you are installing debs, the buildfarm will rebuild for you. The only issue arises if you are using a plugin from older debians and a source checkout of this package, which would be highly frowned upon to begin with.

shaun-edwards:
We would certainly want to see our changes merged into Indigo rather than Jade.  We did our best to make this change backwards compatible and barring the odd situation referenced above, I believe our change is.  That being said, I don't want a mistake on another merge to force this merge.

I've talked to @mikeferguson off list and I believe he has some compelling reasons for his request.  So as not to hijack this PR discussion, I'll contact everyone off list.

isucan:
I have not looked at this in a while, but it sounds like things could be much simpler if we just stick to the LTS branch. I'm inclined to backport changes to indigo and add backward compatibility API, or at least provide support to the mailing list in porting existing kinematics plugins.

kinematics_base/include/moveit/kinematics_base/kinematics_base.h
isucan, line 64: in C++ non-class enums expose the values outside the scope of the enum type, so in order to get something like Foo::Value, using namespaces is the nice thing to do.

isucan, line 64: Brw, another option is to include a prefix in the enum values themselves, which is what we've done more often in MoveIt.

davetcoleman:
I made the mistake of merging https://github.com/ros-planning/moveit_ros/pull/581 into indigo when it should have been in jade, my apologies. I agree with @sachinchitta we should revert it. 

I believe we have these debates about whether we should continue development on older versions of ROS or only on the latest release every time there is a new release. IMHO we should only do development in the latest release (Jade) since that is the point of releases - feature freeze milestones. I as a contributor don't want to have to worry about supporting in backwards compatibility 2 years of old software. Luckily the MoveIt! Jade branches compiles fine with ROS Indigo, and perhaps we should put in effort to maintain that so that we can take advantage of the LTS. But for those people in industry who need stability, they can use older ROS versions of MoveIt! (i.e. Indigo) and know nothing should change except maybe bug fixes.

I think this model of using jade-devel branches provides the most freedom to MoveIt! contributors.

sachinchitta:
Please don't merge this until I have had a chance to test and verify these changes.
Sachin

On Thu, Oct 8, 2015 at 1:12 PM -0700, "Dave Coleman" notifications@github.com wrote:

Merged #231.

PR #233, opened 2015-03-12
fix_type_qualifiers_compiler_warnings

gaspers:
The gcc-compiler warns that the "const" qualifier is ignored when
returning by value. This patch removes unnecessary qualifiers and
compiler warnings emerging from them. It does not have any effect on the
functionality of the code.

### DIFF
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -440,7 +440,7 @@ public:
    *          supported.
    * \return True if the group is supported, false if not.
    */
-  virtual const bool supportsGroup(const moveit::core::JointModelGroup *jmg,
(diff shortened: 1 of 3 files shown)


davetcoleman:
So if I understand correctly, there is no reason to have 'const' when you are returning by value because the memory is being copied over to the caller for it to do as it wants with it?

gaspers:
Yes, that is exactly what happens and why the compiler ignores the qualifier anyway.
For example, with warnings activated (-Wall) the compiler produces the following warning:

```
/opt/ros/indigo/include/moveit/robot_state/robot_state.h:193:65: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
const double getVariablePosition(const std::string &variable) const
```

davetcoleman:
Thanks for the explanation!

PR #242, opened 2015-05-31
API Change of shape_tools

davetcoleman:
Permeating [deprecation of shape_tools](https://github.com/ros-planning/geometric_shapes/pull/32) throughout moveit_*

PR #243, opened 2015-05-31
Indigo devel robot state conversions

davetcoleman:
Includes https://github.com/ros-planning/moveit_core/pull/242

This complements the existing `robotStateToStream` function that I added to write a robot state to file, instead adding a `streamToRobotState` function that reads from a CSV into a robot state

### DIFF
--- a/robot_state/src/conversions.cpp
+++ b/robot_state/src/conversions.cpp
@@ -38,6 +38,7 @@
 #include <moveit/robot_state/conversions.h>
 #include <geometric_shapes/shape_operations.h>
 #include <eigen_conversions/eigen_msg.h>
+#include <boost/lexical_cast.hpp>
 
 namespace moveit
 {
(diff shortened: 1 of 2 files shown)


robot_state/src/conversions.cpp
```diff
@@ -472,3 +472,19 @@ void moveit::core::robotStateToStream(const RobotState& state, std::ostream &out
...
+  std::string cell;
+
+  // For each item/column
+  for (std::size_t i = 0; i < state.getVariableCount(); ++i)
+  {
+    // Get a variable
+    if(!std::getline(lineStream, cell, separator[0]))
+      logError("Missing variable %i", i);
+
+    state.getVariablePositions()[i] = atof(cell.c_str());
```
isucan: It would be better to use boost lexical_cast here, as atof does not really catch errors. I am fine with this as is though, if you don't have time to fix.

davetcoleman: Fixed

davetcoleman:
Ready to be merged.

davetcoleman:
ping

isucan:
Thanks!

PR #247, opened 2015-06-02
Should removing an object from the planning scene also remove entries associated with it in the allowed collision matrix?

mpomarlan:
If yes, see this commit.

A removed object won't affect collisions, of course, but a user who later adds another object with the same name might not expect the acm entries to have persisted.


davetcoleman:
I would actually expect this to persist, and I believe my applications have taken advantage of that fact. In my example, I frequently "recreate" or "repaint" the entire planning scene so as to enable different modes of collision. I want to allow certain parts to collide, but I only set those settings once in the beginning. 

This change might break this assumption that I think I have relied on, and maybe other's too. So since this "feature" has been here for probably a year or two, I say we not change it.

sachinchitta:
Looks like we decided not to merge this - so closing.

PR #250, opened 2015-06-04
Fixed bug in KinematicConstraintSet::decide that makes it evaluate only joint_constraints.

mpomarlan:
Self explanatory. It seems like KinematicConstraintSet::decide intends to go over all kinematic_constraints, rather than merely joint_constraints.


mikeferguson:
Good catch.

PR #251, opened 2015-06-12
Install static libs

garyservin:
Needed for building on [Android](http://wiki.ros.org/android_ndk).


mikeferguson:
LGTM - thanks!

PR #252, opened 2015-06-17
Jade devel sync with indigo

davetcoleman:
We are shifting dev towards jade now, since the new Kinematic changes require it


davetcoleman:
ping - this is just a sync of indigo to jade

davetcoleman:
I've been told by @isucan syncing features in older releases to newer releases is fine to merge, so I'm merging this.

PR #253, opened 2015-06-19
Add object types when requesting attached bodies in planning scene msg.

dornhege:


davetcoleman:
I'm not sure this is the correct place to make this fix. Is there a particular shape type that is not being specified? Mesh perhaps? More details in the PR could be helpful.

I ask because this fix should probably go in one of the sub functions. Perhaps in robot_state::conversions [here](https://github.com/ros-planning/moveit_core/blob/jade-devel/robot_state/src/conversions.cpp#L187)

Or maybe [here](https://github.com/ros-planning/geometric_shapes/blob/ec3c55c55dbaf7846c6f2ac9a20929def459aa38/src/shape_operations.cpp#L374) which is also called in this stack.

I noticed there is likely a bug in geometric_shapes where it doesn't specify a type for meshes, [here](https://github.com/ros-planning/geometric_shapes/blob/ec3c55c55dbaf7846c6f2ac9a20929def459aa38/src/shape_operations.cpp#L428). lmk.

dornhege:
I am also not absolutely sure, where this goes. I mainly replicated the functionality in planning_scene for CollisionObjects (non-attached). I should note that this "type" here is not mesh, primitive, plane or similar, but the object_recognition_msgs/ObjectType (i.e. database + identifier).

It seems to me that internal moveit objects do not have this type anymore and this is manually handled in the planning scene by hasObjectType/getObjectType. This is why it is applied here again and not retrieved directly from the object. If I misinterpreted this, it should obviously be changed.

davetcoleman:
I'm afraid I do not know enough about this part of MoveIt - perhaps @isucan could weight in?

sachinchitta:
@dornhege - Hydro is EOL - do you still want this?

dornhege:
Yes, it still affects all PR2 users that have not done the indigo update, yet.

If someone can confirm that this is a bug, I will also forward-port this to indigo.

dornhege:
#312 now handles this for indigo.

PR #255, opened 2015-08-13
Send Octomap diff instead of whole tree

TheBrewCrew:
This enables Moveit to pass only the diff of the Octomap in a scene diff message when change detection is enabled on the Octree.  This pull request depends on Octomap/octomap#85 to compile.  DO NOT MERGE until the Octomap pull request is accepted.  ros-planning/moveit_ros#602 is the corresponding pull request in Moveit ROS for enabling change detection on the Octomap.

### DIFF
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -49,6 +49,8 @@
 namespace planning_scene
 {
 const std::string PlanningScene::OCTOMAP_NS = "<octomap>";
+const std::string PlanningScene::OCTOMAP_MSG_TYPE = "OcTree";
+const std::string PlanningScene::OCTOMAP_DIFF_MSG_TYPE = "diff(OcTree)";
 const std::string PlanningScene::DEFAULT_SCENE_NAME = "(noname)";
 
 class SceneTransforms : public robot_state::Transforms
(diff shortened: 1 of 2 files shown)


planning_scene/src/planning_scene.cpp
```diff
@@ -861,6 +879,39 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
+  for (octomap::KeyBoolMap::const_iterator it = octree->changedKeysBegin();
+       it != octree->changedKeysEnd() && !(!datastream); ++it)
+  {
+    octomap::OcTreeKey key = it->first;
+    for (int j=0; j<3; j++)
+    {
+      datastream.write((const char*) &key[j], sizeof(unsigned short int));
+    }
+    float value = octree->search(key)->getLogOdds();
+    datastream.write((const char*) &value, sizeof(int));
```
dornhege, line 902: This should probably be `sizeof(float)` instead.

dornhege:
+1 for this.

However, this cannot be accepted as Octomap/octomap#85 will not be released for indigo due to ABI incompatibility. I'll prepare a pull request based on this that only requires Octomap/octomap#103, which should be releasable.

dornhege:
Closing this in favor of ros-planning/moveit_core#274.

PR #259, opened 2015-10-03
Kinematic base updates

jrgnicho:
Reinstated the PR for extending the KinematicsBase interface as described in this PR #231 
@sachinchitta please review.

### DIFF
--- a/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -56,9 +56,13 @@ class RobotState;
 namespace kinematics
 {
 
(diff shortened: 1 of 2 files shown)


davetcoleman:
Is this the same PR that has already been merged into the `jade-devel` branch?

jrgnicho:
@davetcoleman , yes but I had to recreate it since I had deleted the fork from which the PR had been originated.

davetcoleman:
It looks good, but I can't exactly remember - did we agree to also add these changes back into indigo?

jrgnicho:
Yes, I believe so.  In fact I remember being tasked with verifying that the changes didn't break any old (indigo) ikfast implementations.

PR #263, opened 2015-11-13
[cmake] Tell the compiler about FCL include dirs

severin-lemaignan:


davetcoleman:
+1

davetcoleman:
@severin-lemaignan do you think this is related to https://github.com/ros-planning/moveit_core/issues/264 ?

Issue #258, opened 2015-09-18
The CollisionRobotFCL::distanceSelf returns incorrect distance with urdf's that have multiple geometry in a single link.


Levi-Armstrong:
The distanceCallback function in collision_common.cpp needs to check to see if both geometries are associated to the same link. If the below code is add after the "If active components are specified" check along with l1 and l2 being instantiated outside the if statement it solves the issue.

```
// Check if it is an internal link geometry check
if (l1->getName() == l2->getName())
{
  return false;
}
```

mikeferguson:
seems that we have merged fixes last year

PR #265, opened 2015-11-24
Fix issue #258 in indigo-devel

Levi-Armstrong:
This addresses issue #258: When requesting distance information for urdf's that have links with multiple geometries, an incorrect distance is returned. This is a result of the distanceCallback function not checking if both geometries are associated to the same object as done in the [collisionCallback](https://github.com/ros-planning/moveit_core/blob/indigo-devel/collision_detection_fcl/src/collision_common.cpp#L54-L56).


Levi-Armstrong:
@shaun-edwards: FYI

sachinchitta:
+1

shaun-edwards:
@sachinchitta, I assume you noticed this is a PR into indigo.  Should there be a PR into `jade` as well?

Levi-Armstrong:
@sachinchitta, Should I also create a PR into the jade-devel branch?

sachinchitta:
Yes - for all bugs, please open PRs against indigo and jade. All new features go into jade.

PR #272, opened 2016-01-09
Set Joint Model Group Velocities and Accelerations Functions

davetcoleman:
Copies the exact same functionality used for setting and copying position values to a joint model group, but for velocities and accelerations. These are just helper functions for modifying the internal data of a RobotState efficiently


isucan:
I am fine merging this but I purposely did not include such helpers to make the user realize that a lookup is done to find the group, which slows things down. Are you sure you want this in?

isucan:
@sachinchitta @hersh

davetcoleman:
When you say "lookup to find the group" do you mean converting the string name of the group, or do you mean checking if the group's variables are contiguous in memory? I agree its a good idea to remove all functions that take as input joint model groups as string names - all I ever use is `JointModelGroup *group`. I'd like to deprecate those functions throughout all of MoveIt!, really

sachinchitta:
I agree with @isucan on this.

davetcoleman:
I don't understand why this functionality is a bad idea - I'm currently working on a time parametrization algorithm and I'd like to copy a vector of velocities and accelerations into a certain joint model group of the robot. Its easy to do for the positions because @isucan added the function himself. But without this PR how do I do the same for vel & acc? I'm using pointers for the jmg, of course

PR #271, opened 2015-12-15
Remove all trailing whitespace in moveit_core

davetcoleman:
I'm tired of every pull request I make having a bunch of changes all over the file just because my emacs automatically removes trailing whitespace. I'd like to cleanup MoveIt! once and for all of trailing whitespace.

I'm happy to make additional PRs for other branches, repos, as well.

PR #273, opened 2016-01-12
Added maximum acceleration scaling factor

ghost:
This is a change related to ros-planning/moveit_ros#634 and adds a maximum acceleration scaling factor for optionally limiting trajectory acceleration.

### DIFF
--- a/trajectory_processing/include/moveit/trajectory_processing/iterative_time_parameterization.h
+++ b/trajectory_processing/include/moveit/trajectory_processing/iterative_time_parameterization.h
@@ -55,7 +55,8 @@ public:
   ~IterativeParabolicTimeParameterization();
 
   bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory,
-                         const double max_velocity_scaling_factor = 1.0) const;
+                         const double max_velocity_scaling_factor = 1.0,
+                         const double max_acceleration_scaling_factor = 1.0) const;
 
 private:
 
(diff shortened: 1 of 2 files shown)


davetcoleman:
+1 lgtm

could you look into why there is a merge conflict?

ghost:
Yeah, I don't really know what the issue is.  I am fairly new to Git and assumed someone would enlighten me....

davetcoleman:
there are many ways to fix this

make sure you have two remotes added to your .git/config file - github.com/hemes and github.com/ros-planning. then you can create a branch that tracks the upstream version

```
git checkout -b upstream-indigo-devel upstream/indigo-devel
```

switch back to your branch

```
git checkout indigo-devel
```

now merge

```
git merge upstream-indigo-devel
```

to see conflicts

```
git diff
```

commit after resolving the conflicts

ghost:
Nice. That took me longer than it should have but I finally got it.  The issue was with your most recent whitespace changes. Should be good now; thanks for your help!

davetcoleman:
FYI I'm waiting on a +1 from someone else before merging this, since it changes a ROS message in Indigo

dornhege:
In principle this looks good to me, but I can't see a message being changed. There is only one parameter added to the public API, including a default value. Unless we'd want to guarantee ABI compatibility this should be fine.

However, the second merge commit looks odd to me as it contains a lot of (whitespace) changes that aren't visible in the "files changed" tab.

davetcoleman:
This message is being changed:
https://github.com/ros-planning/moveit_msgs/pull/17/files

ghost:
@dornhege Git newcomer here but I believe the commit strangeness is from me updating my fork to include #271, which happened after I forked it initially (and was causing a merge conflict).  Maybe I merged incorrectly somehow?

dornhege:
@davetcoleman I didn't see the dependent pull request. For this one it's a +1 for me also in connection with the message.

@hemes I believe that's from the whitespace changes, so your changes should apply cleanly. It probably only shows because it's a bit forth and back. You could try to remove the merge commit and instead `git rebase upstream/indigo-devel`. I would recommend to backup your repository before you do.

ghost:
@dornhege All right, I think I got it fixed up.

PR #274, opened 2016-01-15
Send Octomap diff instead of whole tree

dornhege:
This is essentially ros-planning/moveit_core#255 by @TheBrewCrew with some minor fixes.

The main reason is that in indigo we can only depend on OctoMap/octomap#103 and not OctoMap/octomap#85. This is handled in 28a6eb4. OctoMap/octomap#103 is already released as 1.6.9 in indigo (currently only shadow-fixed) and thus this can be merged once it is active.

I have left my additions as individual commits, so it is clear for reviewing this, what I changed.
I can also squash them in the original commits before merging, if someone gives me the go-ahead.

The corresponding pull request is moveit_ros is ros-planning/moveit_ros#636.


130s:
Closing in favor of #302

PR #286, opened 2016-04-25
Use libfcl-dev rosdep key in kinetic

### LINKED ISSUE #280: Potential for change in fcl package name
There is some discussion about using the upstream version of the fcl package in ROS K instead of the ROS version (see https://github.com/ros-gbp/fcl-release/issues/1 ). If we change this, then a `kinetic-devel` branch will need to be created so the `package.xml` can be modified with the new name.

scpeters:
Closes #280.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -41,7 +41,7 @@
   <build_depend>eigen</build_depend>
   <build_depend>boost</build_depend>
   <build_depend>assimp</build_depend>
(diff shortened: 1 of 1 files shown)


scpeters:
oh no! I targeted `jade-devel` instead of `kinetic-devel`

scpeters:
We should revert this and retarget to `kinetic-devel`

PR #287, opened 2016-04-27
Revert "Use libfcl-dev rosdep key in kinetic"

scpeters:
Reverts ros-planning/moveit_core#286

it should have targeted `kinetic-devel`


scpeters:
sorry for the trouble

davetcoleman:
oops my bad

PR #289, opened 2016-04-27
[jade] Cherrypick #281

130s:
> getStateAtDurationFromStart would never execute as the check for number of waypoints was inverted

CC @awesomebytes


davetcoleman:
lgtm

PR #291, opened 2016-05-03
Corrected check in getStateAtDurationFromStart

hamalMarino:
The check was returning with 'false' when waypoints were present, instead of the opposite: this prevented the function to correctly execute when a trajectory was set.

Notice that this is already upstream in synaptic v0.7.1 and jade-devel.


awesomebytes:
I answered at the [Merged PR](https://github.com/ros-planning/moveit_core/pull/281#issuecomment-216497628).

rhaschke:
Merged in f1ce912. @130s Do you want to trigger a release fix to fix upstream?

PR #290, opened 2016-04-28
Upgrade to Eigen3 as required in Jade

davetcoleman:

PR #293, opened 2016-05-09
[reopen] Upgrade to Eigen3 as required in Jade

130s:
Replacing #290 

@davetcoleman I [confirmed that this works](https://github.com/130s/moveit_core/pull/1). Tried to open PR against your branch but it seems out of date so I might as well open a new one here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,14 +6,14 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
 endif()
 
 find_package(Boost REQUIRED system filesystem date_time thread iostreams)
-find_package(catkin REQUIRED 
+find_package(catkin REQUIRED
 COMPONENTS
   moveit_msgs
   resource_retriever
   srdfdom
   kdl_parser
   geometric_shapes
-  eigen_stl_containers 
+  eigen_stl_containers
   eigen_conversions
   random_numbers
   visualization_msgs
@@ -22,7 +22,7 @@ COMPONENTS
   cmake_modules
 )
 find_package(console_bridge REQUIRED)
-find_package(Eigen REQUIRED)
+find_package(Eigen3 REQUIRED)
 find_package(octomap REQUIRED)
 find_package(urdfdom REQUIRED)
 find_package(urdfdom_headers REQUIRED)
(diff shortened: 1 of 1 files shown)


130s:
Ping @davetcoleman or anyone who's interested. Thanks!

rhaschke:
@130s Is https://goo.gl/hNNKD1 the recommended way to [fix issues with finding Eigen3 on OSX](https://github.com/ros-perception/perception_pcl/pull/108)? From my understanding this issue is not yet resolved, is it?

CMakeLists.txt
```diff
@@ -22,7 +22,10 @@ COMPONENTS
   cmake_modules
 )
 find_package(console_bridge REQUIRED)
-find_package(Eigen REQUIRED)
+# Eigen3 hack for Jade http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules
+find_package(Eigen3)
+find_package(PkgConfig)
+pkg_search_module(Eigen3 REQUIRED eigen3)
```
davetcoleman: I don't understand how this hack is the same as the one in the link you [reference](http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules), they look pretty different to me.

130s:
@rhaschke @davetcoleman yeah I'm not sure whether this PR that got a hint from https://goo.gl/hNNKD1 that uses `pkg_config` is the right solution, particularly on ROS buildfarm where I know some people raised issues using pkg_config. I just confirmed that the build passes on Travis. Sorry but I don't have other idea to move this on. I asked [a question](http://answers.ros.org/question/233344/jade-eigen3-not-found-build-error/) but got no answer yet.

rhaschke:
I looked into that issue a little bit deeper: One issue is that `CMAKE_MODULE_PATH` is not set. (Should it be set by `catkin tools`?). But on Trusty (used by Travis CI), there is only a `FindEigen3.cmake` (in `/usr/share/cmake-2.8/Modules`) requiring the `CMAKE_MODULE_PATH` to point there. Eventually, on Xenial this looks different.

On the other hand, finding `Eigen3` via `pkg-config` does work. The fallback to cmake_modules suggested in the [reference](http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules), is not needed anymore then (actually the latter [finds Eigen via `pkg-config` as well](https://github.com/ros/cmake_modules/blob/0.4-devel/cmake/Modules/FindEigen.cmake#L58-L59).)

However, case matters! The line should be changed to
`pkg_search_module(EIGEN3 REQUIRED eigen3)`
filling the variable `EIGEN3_INCLUDE_DIRS` - in contrast to `EIGEN_INCLUDE_DIR` filled by the old `FindEigen.cmake` of `cmake_modules`.

Anyway, in the moveit_core package, Eigen3 doesn't need to be pulled in explicitly, because it catkin-depends on `eigen_stl_containers` and `eigen_conversions`, which pulls in Eigen as well. That's why it worked, no matter which variables were set or not.

I rebased @130s' branch to `jade-devel` as branch [`jade-devel-eigen3`](https://github.com/ros-planning/moveit_core/compare/jade-devel-eigen3?expand=1). Feel free to continue there. I guess we should squash before merging ;-)

130s:
@rhaschke thanks for the deeper look. I cherry-picked your commits into this PR, and squashed a little bit (I still left 3 commits to show all the committers who contributed).

rhaschke:
LGTM.

rhaschke:
@130s Can you adapt the other MoveIt repos in a similar fashion?

PR #295, opened 2016-05-25
Fix fcl implementation so the fcl::CollisionObject's are not created every time a request is made.

Levi-Armstrong:
Every time a collision or distance request is made all of the fcl::CollisionObjects are recreated shown [here](https://github.com/ros-planning/moveit_core/blob/indigo-devel/collision_detection_fcl/src/collision_robot_fcl.cpp#L76-L83). During the creation of the fcl::CollisionObject it computes the AABB of the passed geometry as shown [here](https://github.com/flexible-collision-library/fcl/blob/bdb34882ab2b5cdc14aee56bd65ab5723a15a589/include/fcl/collision_object.h#L171-L176). This calculation is very expensive because it has to walk the mesh twice as shown [here](https://github.com/flexible-collision-library/fcl/blob/bdb34882ab2b5cdc14aee56bd65ab5723a15a589/src/BVH/BVH_model.cpp#L857-L877). The calculation of the geometries AABB is only required once and then it can easily be update by providing the TF using [CollisionObject::setTransform](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/collision_object.h#L283-L286) followed by calling the [CollisionObject::updateAABB](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/collision_object.h#L201-L214) which simply re-sizes the AABB based on the transform as shown [here](https://github.com/Levi-Armstrong/moveit_core/blob/improveCollRobotIndigo_V2/collision_detection_fcl/src/collision_robot_fcl.cpp#L92-L96). It is also seen here that I am creating a copy of the cached fcl::CollisionObject which is to maintain thread safe.

I performed some baseline testing. The testing was performed using a joint interpolated planner and performing 100 planning requests using the same start and goal then taking the average of the time it took to calculate the trajectory where each performed a 100 collision checks. I performed this using both methods where I changes the meshing of a single object in the urdf to where the total triangle count was 10K, 500K and 10,000K. The performance improvements are shown below and they are significant.

| Triangle Count (K) | indigo-devel (sec) | New Method (sec) | % Improvement | Speed Improvement |
| --- | :-: | :-: | :-: | --: |
| 10 | 0.0129 | 0.00229 | 82.25% | 5.6X |
| 500 | 0.1459 | 0.00229 | 98.43% | 63.7X |
| 10,000 | 3.1449 | 0.00263 | 99.92% | 1195.8X |

### DIFF
--- a/collision_detection_fcl/src/collision_robot_fcl.cpp
+++ b/collision_detection_fcl/src/collision_robot_fcl.cpp
@@ -56,7 +56,7 @@ collision_detection::CollisionRobotFCL::CollisionRobotFCL(const robot_model::Rob
         // Need to store the FCL object so the AABB does not get recreated every time.
         // Every time this object is created, g->computeLocalAABB() is called  which is
         // very expensive and should only be calculated once. To update the AABB, use the
-        // collObj->setTsetTransform and then call collObj->computeAABB() to transform the AABB.
+        // collObj->setTransform and then call collObj->computeAABB() to transform the AABB.
         fcl_objs_[index] = FCLCollisionObjectConstPtr(new fcl::CollisionObject(g->collision_geometry_));
       }
       else
@@ -84,14 +84,14 @@ void collision_detection::CollisionRobotFCL::getAttachedBodyObjects(const robot_
 void collision_detection::CollisionRobotFCL::constructFCLObject(const robot_state::RobotState &state, FCLObject &fcl_obj) const
 {
   fcl_obj.collision_objects_.reserve(geoms_.size());
-  fcl::Transform3f tf;
+  fcl::Transform3f fcl_tf;
 
   for (std::size_t i = 0 ; i < geoms_.size() ; ++i)
     if (geoms_[i] && geoms_[i]->collision_geometry_)
     {
-      transform2fcl(state.getCollisionBodyTransform(geoms_[i]->collision_geometry_data_->ptr.link, geoms_[i]->collision_geometry_data_->shape_index), tf);
+      transform2fcl(state.getCollisionBodyTransform(geoms_[i]->collision_geometry_data_->ptr.link, geoms_[i]->collision_geometry_data_->shape_index), fcl_tf);
       fcl::CollisionObject *collObj = new fcl::CollisionObject(*fcl_objs_[i]);
-      collObj->setTransform(tf);
+      collObj->setTransform(fcl_tf);
       collObj->computeAABB();
       fcl_obj.collision_objects_.push_back(FCLCollisionObjectPtr(collObj));
     }
@@ -107,8 +107,8 @@ void collision_detection::CollisionRobotFCL::constructFCLObject(const robot_stat
     for (std::size_t k = 0 ; k < objs.size() ; ++k)
       if (objs[k]->collision_geometry_)
       {
-        transform2fcl(ab_t[k], tf);
-        fcl_obj.collision_objects_.push_back(FCLCollisionObjectPtr(new fcl::CollisionObject(objs[k]->collision_geometry_, tf)));
+        transform2fcl(ab_t[k], fcl_tf);
+        fcl_obj.collision_objects_.push_back(FCLCollisionObjectPtr(new fcl::CollisionObject(objs[k]->collision_geometry_, fcl_tf)));
         // we copy the shared ptr to the CollisionGeometryData, as this is not stored by the class itself,
         // and would be destroyed when objs goes out of scope.
         fcl_obj.collision_geometry_.push_back(objs[k]);
--- a/collision_detection_fcl/src/collision_robot_fcl.cpp
+++ b/collision_detection_fcl/src/collision_robot_fcl.cpp
@@ -40,14 +40,25 @@ collision_detection::CollisionRobotFCL::CollisionRobotFCL(const robot_model::Rob
   : CollisionRobot(model, padding, scale)
 {
   const std::vector<const robot_model::LinkModel*>& links = robot_model_->getLinkModelsWithCollisionGeometry();
+  std::size_t index;
   geoms_.resize(robot_model_->getLinkGeometryCount());
+  fcl_objs_.resize(robot_model_->getLinkGeometryCount());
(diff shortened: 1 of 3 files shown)


Jmeyer1292:
This change allows us to actually take advantage of FCL's broad phase collision detection. We no longer pay a significant price for complex models that are not near collision during runtime. Profiling of our applications has shown many spend 99% of their time in collision checking. I encourage you to run any of your moveit applications that may take into account collisions under something like `valgrind --tool=callgrind` and see for yourself.

It would be great if the community at large could inspect this change, test it out, and lend their brains to the implications to related factors such as thread-safety.

I'm tagging maintainers: @isucan @mikeferguson @acornacorn 
and frequent contributor: @davetcoleman This may be relevant to your work as well.

collision_detection_fcl/src/collision_robot_fcl.cpp
```diff
@@ -40,14 +40,25 @@ collision_detection::CollisionRobotFCL::CollisionRobotFCL(const robot_model::Rob
...
       if (g)
-        geoms_[links[i]->getFirstCollisionBodyTransformIndex() + j] = g;
+      {
+        index = links[i]->getFirstCollisionBodyTransformIndex() + j;
+        geoms_[index] = g;
+
+        // Need to store the FCL object so the AABB does not get recreated every time.
+        // Every time this object is created, g->computeLocalAABB() is called  which is
+        // very expensive and should only be calculated once. To update the AABB, use the
+        // collObj->setTsetTransform and then call collObj->computeAABB() to transform the AABB.
```
davetcoleman: Misspelling `setTsetTransform`?

collision_detection_fcl/src/collision_robot_fcl.cpp
```diff
@@ -72,16 +84,19 @@ void collision_detection::CollisionRobotFCL::getAttachedBodyObjects(const robot_
 void collision_detection::CollisionRobotFCL::constructFCLObject(const robot_state::RobotState &state, FCLObject &fcl_obj) const
 {
   fcl_obj.collision_objects_.reserve(geoms_.size());
+  fcl::Transform3f tf;
```
davetcoleman: I don't like the var name `tf` because it can confuse ROS users used to the `TF` package

davetcoleman:
This is a great find! I will begin to test this on my end.

FYI the following link is not viewable for me:
https://raesgit.datasys.swri.edu/larmstrong/moveit_core/blob/improveCollRobotIndigo/collision_detection_fcl/src/collision_robot_fcl.cpp#L90-L98

collision_detection_fcl/src/collision_robot_fcl.cpp
Levi-Armstrong: Do you have preference on the name or just anything but tf?

Levi-Armstrong:
> FYI the following link is not viewable for me:
> https://raesgit.datasys.swri.edu/larmstrong/moveit_core/blob/improveCollRobotIndigo/collision_detection_fcl/src/collision_robot_fcl.cpp#L90-L98

The link has been fixed.

collision_detection_fcl/src/collision_robot_fcl.cpp
davetcoleman: no preference, if you want to leave it tf its not a big deal

collision_detection_fcl/src/collision_robot_fcl.cpp
davetcoleman: ping

davetcoleman:
I just ran this using the default Hilgendorf (two UR5's and a base stand) with a simple wall obstacle and floor obstacle around it. I randomly sampled 10,000 joint configurations and asked MoveIt! if the state was collision free. I unfortunately only saw a 12.6% improvement in the speed of this task:

Before this commit: 21.0 sec
After this commit: 18.4 sec

Is my significantly test different than yours? Note I'm compiled in debug mode and using FCL through OMPL.

hilgendorf_with_wall

Levi-Armstrong:
The reason may be that the performance is driven by the number of triangles in the urdf collision geometries so the triangle count may be significantly lower and 10,000. You may try adding an object with around 500,000 triangles and you should see a big difference.

collision_detection_fcl/src/collision_robot_fcl.cpp
Levi-Armstrong: Sorry, I have fixed the misspelling, but I have not pushed the changes yet. I was waiting to see if there were any addition modification required.

Jmeyer1292:
@davetcoleman I would also add that:
1. Levi's numbers are sort of 'best case scenario' as they are mostly collision checks against objects whose bounding boxes do not intersect (Hence the flat cost of collision checks despite growing mesh size).
2. All tests were done in Release mode. I doubt it makes a huge difference here, but making statements about the performance of code should always be done with optimization enabled.

Levi-Armstrong:
@davetcoleman, I made the requested changes. I can also fixup this change so there is only one commit if you like.

davetcoleman:
Since two of you are in agreement on the PR and since I have been running it successfully for a week I'll give it the green light, thanks!

Do you by chance have a re-usable benchmark setup for collision checking that we should save for future use? Not sure where the best place to put it would be but seems like a useful tool to have.

Also, do you mind sharing a summary of this great improvement on the [mailing list](https://groups.google.com/forum/#!forum/moveit-users) to help us drum up support for MoveIt! development?

Levi-Armstrong:
> Do you by chance have a re-usable benchmark setup for collision checking that we should save for future use? Not sure where the best place to put it would be but seems like a useful tool to have.

There will be one available today or tomorrow as part of the [industrial_moveit repository](https://github.com/ros-industrial/industrial_moveit). This is where we have been make enhancements to moveit so we have a benchmarking package.

Levi-Armstrong:
@davetcoleman Should I create a pull request against jade-devel or will you cherry pick it?

davetcoleman:
I would welcome you to create a quick PR, I'm actively working on Docker images to make it easier for me to do things like cherry-pick

PR #297, opened 2016-06-14
planning_scene updates: expose success state to caller

v4hn:
This is required to get the information back for the
ApplyPlanningSceneService.

This is required for https://github.com/ros-planning/moveit_ros/pull/686

As it turns out, according to abi-compliance-checker these changes
are _also_ ABI compliant... I didn't expect that.

### DIFF
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -1133,8 +1133,10 @@ void planning_scene::PlanningScene::decoupleParent()
   parent_.reset();
 }
 
-void planning_scene::PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::PlanningScene &scene_msg)
+bool planning_scene::PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::PlanningScene &scene_msg)
 {
+  bool result = true;
+
   logDebug("moveit.planning_scene: Adding planning scene diff");
   if (!scene_msg.name.empty())
     name_ = scene_msg.name;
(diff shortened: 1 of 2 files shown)


planning_scene/src/planning_scene.cpp
```diff
@@ -1185,14 +1187,16 @@ void planning_scene::PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::P
 
   // process collision object updates
   for (std::size_t i = 0 ; i < scene_msg.world.collision_objects.size() ; ++i)
-    processCollisionObjectMsg(scene_msg.world.collision_objects[i]);
+    ret &= processCollisionObjectMsg(scene_msg.world.collision_objects[i]);
```
davetcoleman: I'm wondering if this is out of style for MoveIt!'s code base - we have almost no `&=` in the code base (except in the rviz plugin, twice) and I found it hard to understand at first. Wouldn't `+=` accomplish the same thing but be easier to quickly understand? Its probably just me being ignorant but I'm not used to seeing it

planning_scene/src/planning_scene.cpp
```diff
@@ -1133,8 +1133,10 @@ void planning_scene::PlanningScene::decoupleParent()
   parent_.reset();
 }
 
-void planning_scene::PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::PlanningScene &scene_msg)
+bool planning_scene::PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::PlanningScene &scene_msg)
 {
+  bool ret = true;
```
davetcoleman: Minor but, MoveIt!'s code style is to write out variable names in full form as much as possible. `bool result` is very common throughout the code base, so I think this variable should also be called `result`, thanks!

v4hn: `&=` is a standard C operator, I don't see why we shouldn't use it.
`+=` might even work, but how do you add boolean values?
Now _that_ doesn't make sense to me :-)

The only alternatives I see to `a&=b` are `a=a&b` or `a=a && b`.
If you favor one of them, I can change it. I don't really care...

btw, I just changed the variable name.

130s:
Looks like this is already cherry-picked into `jade-devel` https://github.com/ros-planning/moveit_core/commit/702c3e75b7f1cf79c2096d8cf5466f614b830a07 w/o PR.

v4hn:
Yes. When I asked about a cherry-pick policy dave replied
that it's fine to pick merged commits to newer branches without PR [here](https://github.com/ros-planning/moveit_msgs/issues/22#issuecomment-226026842).
Would you prefer to have a PR for _each_ cherry-pick @130s ?

130s:
@v4hn Yeah I think so,,,although I know it's cumbersome. 
As I commented https://github.com/ros-planning/moveit_core/pull/301#issuecomment-229156460, currently I don't know an easy, cleaner way to find commits that need synced to other branches, and going through PRs is robust enough I think. Suggestion is much welcomed.

v4hn:
https://github.com/ros-planning/moveit_core/pull/301#issuecomment-229187056
for reference. Let's continue this over there.

davetcoleman:
I still think if you _just_ had a PR merged and you want to reflect it in a newer branch, its no big deal to cherry-pick and commit it without a commit. I say its fine because its what I've seen others do here and it makes sense to me. I created those PRs for syncing in the past day because they were old and messy and potential error-prone. I think we should just do a better job of always keeping the branches in sync

PR #300, opened 2016-06-27
Jade devel synced

davetcoleman:
Sync indigo commits to jade

@v4hn my emacs didn't know what to do on the CHANGELOG.rst encoding, I told it not to change anything but your name still looks mangled. Any advice?


v4hn:
I do not particularly care. I don't expect tools to handle utf-8 correctly for the next 10 years at least. :-)
If you want to have a clean encoding in the file, either replace it by my nick or rewrite the umlaut to 'oe'.

Good idea to clean this up. This includes at least two ABI breaks, but I do not think this is too interesting given that the changes already made it into indigo... Other than that +1.

davetcoleman:
I changed your name to "Goerner", sorry our puny computers can't handle utf-8

PR #301, opened 2016-06-28
Jade devel synced

davetcoleman:
Replaces https://github.com/ros-planning/moveit_core/pull/300 because  I guess the git history was too messy. Here I attempted to manually cherry-pick commits from the indigo-devel branch, I'm pretty sure I got them all.


130s:
+1. Looks exhaustive after tracking PRs (see below).

[PRs merged since May 18 (date the last release was made on indigo) into indigo-devel](https://github.com/ros-planning/moveit_core/pulls?utf8=%E2%9C%93&q=base%3Aindigo-devel%20is%3Apr%20is%3Aclosed%20sort%3Aupdated-desc%20merged%3A%3E%3D2016-05-18%20):
- [x] https://github.com/ros-planning/moveit_core/pull/297 or https://github.com/ros-planning/moveit_core/pull/297/commits/0322f52193dcf535ae87191e297955e6bcf632ad --> Done at https://github.com/ros-planning/moveit_core/pull/297#issuecomment-229153262 w/o PR.
- https://github.com/ros-planning/moveit_core/pull/295 or
  - [x] https://github.com/ros-planning/moveit_core/pull/295/commits/456a25956c13ac7e90f024c4bcda3166334b3906
  - [x] https://github.com/ros-planning/moveit_core/pull/295/commits/a796f68818110b37331cd5dace5a83adc7163de3

Also the following is not obvious from merged PRs but I know it's probably necessary:
- [x] https://github.com/ros-planning/moveit_core/pull/291 or https://github.com/ros-planning/moveit_core/commit/f1ce912f2c216de9796c95dd28f9ddd13bb8949f --> done at https://github.com/ros-planning/moveit_core/commit/6946c06833e109182f663e34ef31deec1218ff40 w/o PR.

---

(We might want to find auto/semi-automated way to keep track commits to sync between branches. Also, cherry-pick should be done via PR only, no direct commit, to make tracking history easier).

davetcoleman:
Can you merge it then? There is nothing new here, just a cherry-pick

v4hn:
On Tue, Jun 28, 2016 at 12:29:49PM -0700, Isaac I.Y. Saito wrote:

> (We might want to find auto/semi-automated way to keep track commits to sync between branches.

`git cherry -v jade-devel indigo-devel | grep '^+'` lists all commits that are part of
indigo-devel, but not present in jade-devel. Is that what you are looking for?

> Also, cherry-pick should be done via PR only, no direct commit, to make tracking history easier).

Until now I considered cherry-picks to be much better for a simple commit history
and keeping track of the repository. Merge requests usually make the history non-linear
and the commit graphs are pretty confusing. Not to mention resulting merge-commits -
unless the maintainer remembers to use github's squash-merge (which is actually a fancy cherry-pick).
What makes you say pull requests help to track history?

130s:
> `git cherry -v jade-devel indigo-devel | grep '^+'` lists all commits that are part of indigo-devel, but not present in jade-devel

Looks useful!
And I see your point.

> What makes you say pull requests help to track history?

Not all PRs need to be synchronized between repos; for example from other repo, https://github.com/ros-planning/warehouse_ros/pull/32 is necessary only for jade-devel or higher but that is not clear from the commit message. I think PR creators tend to explain it in the PR message, and IMO it's fine to not specify which branch the commit should be applied to. Thus I think opening PR for cherry-pick commits still works better (far from ideal though) in terms of avoiding over-cherrypicking. Thoughts?

v4hn:
On Wed, Jun 29, 2016 at 08:50:54AM -0700, Isaac I.Y. Saito wrote:

> for example from other repo, https://github.com/ros-planning/warehouse_ros/pull/32 is necessary
> only for jade-devel or higher but that is not clear from the commit message.
> I think PR creators tend to explain it in the PR message,
> and IMO it's fine to not specify which branch the commit should be applied to.
> Thus I think opening PR for cherry-pick commits still works better (far from ideal though) in terms of avoiding over-cherrypicking. Thoughts?

In my experience devs open up pull-requests against whichever branch they are actually using.
I would prefer to see the discussion on which branches a request should go to _as part of the pull-request_.
A simple "+1 to merge into indigo-devel, jade-devel and kinetic-devel" would be fine for many cases.
The developer who merges the change later on can readily cherry-pick the change to other branches afterward
and already got confirmation to cherry-pick.

For complex / ABI breaking changes that have to be back-/forward ported manually between branches
and can not simply be cherry-picked, I also believe that new pull requests should be created.

davetcoleman:
+1 to specifying which branches the reviewer thinks it should go to.

130s:
Ok, so it's the merger's responsibility to cherry-pick toward the agreed branch(es)? If so +1 as long as the cherry-pick happens right away after the merge (otherwise the same issue remains unsolved).

PR #302, opened 2016-06-28
[reopen  #274] Send Octomap diff instead of whole tree

130s:
> This enables Moveit to pass only the diff of the Octomap in a scene diff message when change detection is enabled on the Octree. This pull request depends on OctoMap/octomap#85 to compile. DO NOT MERGE until the Octomap pull request is accepted. ros-planning/moveit_ros#602 is the corresponding pull request in Moveit ROS for enabling change detection on the Octomap.

@dornhege I've rebased #274 to include updates (esp. CI config for testing). Thanks.

### DIFF
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -858,11 +858,12 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
       boost::shared_ptr<const octomap::OcTree> octree = o->octree;
       if (scene_msg.is_diff == true && octree->isChangeDetectionEnabled())
       {
-        int expected_size_diff = sizeof(int)+octree->numChangesDetected()*((3*sizeof(unsigned short int))+sizeof(float));
+        int num_changes = octree->numChangesDetected();
+        int expected_size_diff = sizeof(int)+num_changes*((3*sizeof(unsigned short int))+sizeof(float));
         int expected_size_tree = octree->size()*(sizeof(float)+sizeof(char));
         if (expected_size_diff > expected_size_tree)
         {
-          logInform("Cheaper to send tree instead of diff by %i bytes with %i changes", expected_size_diff-expected_size_tree, octree->numChangesDetected());
+          logInform("Cheaper to send tree instead of diff by %i bytes with %i changes", expected_size_diff-expected_size_tree, num_changes);
           octomap_msgs::fullMapToMsg(*octree, scene_msg.world.octomap.octomap);
           if(scene_msg.world.octomap.octomap.id != OCTOMAP_MSG_TYPE) {
             logWarn("fullMapToMsg produced unexpected octomap type: %s",
@@ -900,8 +901,10 @@ bool planning_scene::PlanningScene::getPlanningSceneMsgOctomapDiff(boost::shared
   datastream.write((const char*) &num_changes, sizeof(int));
   //logInform("Octomap diff has %i changes", num_changes);
 
-  for (octomap::KeyBoolMap::const_iterator it = octree->changedKeysBegin();
-       it != octree->changedKeysEnd() && !(!datastream); ++it)
+  // this is safe as long as we only use the const_iterators for changedKeys
+  octomap::OcTree* octreeNonConst = const_cast<octomap::OcTree*>(octree.get());
+  for (octomap::KeyBoolMap::const_iterator it = octreeNonConst->changedKeysBegin();
+       it != octreeNonConst->changedKeysEnd() && !(!datastream); ++it)
   {
     octomap::OcTreeKey key = it->first;
     for (int j=0; j<3; j++)
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -1245,32 +1245,6 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
     return setPlanningSceneMsg(scene_msg);
 }
 
-void planning_scene::PlanningScene::processOctomapMsg(const octomap_msgs::Octomap &map)
-{
-  // each octomap replaces any previous one
-  world_->removeObject(OCTOMAP_NS);
-
-  if (map.data.empty())
-    return;
-
-  if (map.id != "OcTree")
-  {
-    logError("Received octomap is of type '%s' but type 'OcTree' is expected.", map.id.c_str());
-    return;
-  }
-
-  boost::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map)));
-  if (!map.header.frame_id.empty())
-  {
-    const Eigen::Affine3d &t = getTransforms().getTransform(map.header.frame_id);
-    world_->addToObject(OCTOMAP_NS, shapes::ShapeConstPtr(new shapes::OcTree(om)), t);
-  }
-  else
-  {
-    world_->addToObject(OCTOMAP_NS, shapes::ShapeConstPtr(new shapes::OcTree(om)), Eigen::Affine3d::Identity());
-  }
-}
-
 void planning_scene::PlanningScene::removeAllCollisionObjects()
 {
   const std::vector<std::string> &object_ids = world_->getObjectIds();
@@ -1281,24 +1255,44 @@ void planning_scene::PlanningScene::removeAllCollisionObjects()
 
 void planning_scene::PlanningScene::processOctomapMsg(const octomap_msgs::OctomapWithPose &map)
 {
-  // each octomap replaces any previous one
-  world_->removeObject(OCTOMAP_NS);
-
-  if (map.octomap.data.empty())
-    return;
-
-  if (map.octomap.id != "OcTree")
-  {
-    logError("Received octomap is of type '%s' but type 'OcTree' is expected.", map.octomap.id.c_str());
-    return;
-  }
-
-  boost::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map.octomap)));
   const Eigen::Affine3d &t = getTransforms().getTransform(map.header.frame_id);
   Eigen::Affine3d p;
   tf::poseMsgToEigen(map.origin, p);
   p = t * p;
-  world_->addToObject(OCTOMAP_NS, shapes::ShapeConstPtr(new shapes::OcTree(om)), p);
+  processOctomapMsg(map.octomap, p);
+}
+
+void planning_scene::PlanningScene::processOctomapMsg(const octomap_msgs::Octomap &map)
+{
+  if (!map.header.frame_id.empty())
+  {
+    const Eigen::Affine3d &t = getTransforms().getTransform(map.header.frame_id);
+    processOctomapMsg(map, t);
+  }
+  else
+  {
+    processOctomapMsg(map, Eigen::Affine3d::Identity());
+  }
+}
+
+void planning_scene::PlanningScene::processOctomapMsg(const octomap_msgs::Octomap &map, const Eigen::Affine3d &t)
+{
+  if (map.id.empty())
+  {
+    world_->removeObject(OCTOMAP_NS);
+  }
+  else if (map.id == "OcTree")
+  {
+    world_->removeObject(OCTOMAP_NS); // Octomap replaces any previous one
+    if (map.data.empty())
+      return;
+    boost::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map)));
+    world_->addToObject(OCTOMAP_NS, shapes::ShapeConstPtr(new shapes::OcTree(om)), t);
+  }
+  else
+  {
+    logError("Received Octomap is of unknown type '%s'", map.id.c_str());
+  }
 }
 
 void planning_scene::PlanningScene::processOctomapPtr(const boost::shared_ptr<const octomap::OcTree> &octree, const Eigen::Affine3d &t)
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -899,7 +899,7 @@ bool planning_scene::PlanningScene::getPlanningSceneMsgOctomapDiff(boost::shared
       datastream.write((const char*) &key[j], sizeof(unsigned short int));
     }
     float value = octree->search(key)->getLogOdds();
-    datastream.write((const char*) &value, sizeof(int));
+    datastream.write((const char*) &value, sizeof(float));
   }
 
   if (!datastream)
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -49,6 +49,8 @@
 namespace planning_scene
 {
 const std::string PlanningScene::OCTOMAP_NS = "<octomap>";
+const std::string PlanningScene::OCTOMAP_MSG_TYPE = "OcTree";
+const std::string PlanningScene::OCTOMAP_DIFF_MSG_TYPE = "diff(OcTree)";
 const std::string PlanningScene::DEFAULT_SCENE_NAME = "(noname)";
 
 class SceneTransforms : public robot_state::Transforms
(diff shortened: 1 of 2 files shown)


130s:
Comment from the original PR:

This is essentially ros-planning/moveit_core#255 by @TheBrewCrew with some minor fixes.

The main reason is that in indigo we can only depend on OctoMap/octomap#103 and not OctoMap/octomap#85. This is handled in 28a6eb4. OctoMap/octomap#103 is already released as 1.6.9 in indigo (currently only shadow-fixed) and thus this can be merged once it is active.

I have left my additions as individual commits, so it is clear for reviewing this, what I changed.
I can also squash them in the original commits before merging, if someone gives me the go-ahead.

The corresponding pull request is moveit_ros is ros-planning/moveit_ros#636.

dornhege:
+1
I've tested this on the current indigo head in connection with ros-planning/moveit_ros#636 and it still works. I shouldn't merge this though as I did some of the changes on #274.

planning_scene/src/planning_scene.cpp
```diff
@@ -853,14 +855,76 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
+      if (scene_msg.is_diff == true && octree->isChangeDetectionEnabled())
+      {
+        int num_changes = octree->numChangesDetected();
+        int expected_size_diff = sizeof(int)+num_changes*((3*sizeof(unsigned short int))+sizeof(float));
+        int expected_size_tree = octree->size()*(sizeof(float)+sizeof(char));
+        if (expected_size_diff > expected_size_tree)
+        {
+          logInform("Cheaper to send tree instead of diff by %i bytes with %i changes", expected_size_diff-expected_size_tree, num_changes);
+          octomap_msgs::fullMapToMsg(*octree, scene_msg.world.octomap.octomap);
+          if(scene_msg.world.octomap.octomap.id != OCTOMAP_MSG_TYPE) {
```
davetcoleman, line 868: curly brackets on new line

planning_scene/src/planning_scene.cpp
```diff
@@ -853,14 +855,76 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
+        }
+      }
       tf::poseEigenToMsg(map->shape_poses_[0], scene_msg.world.octomap.origin);
     }
     else
       logError("Unexpected number of shapes in octomap collision object. Not including '%s' object", OCTOMAP_NS.c_str());
   }
 }
 
+bool planning_scene::PlanningScene::getPlanningSceneMsgOctomapDiff(boost::shared_ptr<const octomap::OcTree> octree, octomap_msgs::Octomap &msg) const
```
davetcoleman, line 893: line too long

planning_scene/src/planning_scene.cpp
```diff
@@ -853,14 +855,76 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
+  datastream.write((const char*) &num_changes, sizeof(int));
+  //logInform("Octomap diff has %i changes", num_changes);
+
+  // this is safe as long as we only use the const_iterators for changedKeys
+  octomap::OcTree* octreeNonConst = const_cast<octomap::OcTree*>(octree.get());
+  for (octomap::KeyBoolMap::const_iterator it = octreeNonConst->changedKeysBegin();
+       it != octreeNonConst->changedKeysEnd() && !(!datastream); ++it)
+  {
+    octomap::OcTreeKey key = it->first;
+    for (int j=0; j<3; j++)
```
davetcoleman, line 910: spaces between symbols

planning_scene/src/planning_scene.cpp
```diff
@@ -1245,60 +1309,101 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
...
+  datastream.read((char*) &num_changes, sizeof(int));
+  //logInform("Octomap diff has %i changes", num_changes);
+  int expected_size = sizeof(int)+num_changes*((3*sizeof(unsigned short int))+sizeof(float));
+  if (expected_size > msg.data.size())
   {
-    logError("Received octomap is of type '%s' but type 'OcTree' is expected.", map.octomap.id.c_str());
+    logError("Did not receive enough data for specified diff size: %i bytes expected, %i received", expected_size, msg.data.size());
     return;
   }
+  if(expected_size < msg.data.size()) {
```
davetcoleman, line 1390: bracket break

planning_scene/src/planning_scene.cpp
```diff
@@ -1245,60 +1309,101 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
...
+      logWarn("Got more data than expected (%zu > %d)", msg.data.size(), expected_size);
+  }
 
-  boost::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map.octomap)));
-  const Eigen::Affine3d &t = getTransforms().getTransform(map.header.frame_id);
-  Eigen::Affine3d p;
-  tf::poseMsgToEigen(map.origin, p);
-  p = t * p;
-  world_->addToObject(OCTOMAP_NS, shapes::ShapeConstPtr(new shapes::OcTree(om)), p);
+  for (int i=0; i<num_changes && !(!datastream); ++i)
```
davetcoleman, line 1394: spaces

davetcoleman:
I wish we would all move away from deep changes like this in an LTS (indigo) release...

planning_scene/src/planning_scene.cpp
v4hn, line 893: Do we have a standard concerning maximum line length?
Personally I prefer to have even long function signatures in a single line, to support `grep` users.
But I see the disadvantages of that too..

planning_scene/src/planning_scene.cpp
```diff
@@ -853,14 +855,76 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
-      octomap_msgs::fullMapToMsg(*o->octree, scene_msg.world.octomap.octomap);
+      boost::shared_ptr<const octomap::OcTree> octree = o->octree;
+      if (scene_msg.is_diff == true && octree->isChangeDetectionEnabled())
+      {
+        int num_changes = octree->numChangesDetected();
+        int expected_size_diff = sizeof(int)+num_changes*((3*sizeof(unsigned short int))+sizeof(float));
+        int expected_size_tree = octree->size()*(sizeof(float)+sizeof(char));
+        if (expected_size_diff > expected_size_tree)
+        {
+          logInform("Cheaper to send tree instead of diff by %i bytes with %i changes", expected_size_diff-expected_size_tree, num_changes);
```
v4hn, line 866: should this really be an inform? I would prefer debug.

planning_scene/src/planning_scene.cpp
```diff
@@ -853,14 +855,76 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
+          getPlanningSceneMsgOctomapDiff(octree, scene_msg.world.octomap.octomap);
+        }
+      }
+      else
+      {
+        octomap_msgs::fullMapToMsg(*octree, scene_msg.world.octomap.octomap);
+        if(scene_msg.world.octomap.octomap.id != OCTOMAP_MSG_TYPE) {
+          logWarn("fullMapToMsg produced unexpected octomap type: %s",
+                  scene_msg.world.octomap.octomap.id.c_str());
+        }
```
v4hn, line 884: This looks like something that was needed during development, but bloats the final result.
Is this still needed? In my opinion, this check should be removed.

planning_scene/src/planning_scene.cpp
```diff
@@ -853,14 +855,76 @@ void planning_scene::PlanningScene::getPlanningSceneMsgOctomap(moveit_msgs::Plan
...
+        int expected_size_diff = sizeof(int)+num_changes*((3*sizeof(unsigned short int))+sizeof(float));
+        int expected_size_tree = octree->size()*(sizeof(float)+sizeof(char));
+        if (expected_size_diff > expected_size_tree)
+        {
+          logInform("Cheaper to send tree instead of diff by %i bytes with %i changes", expected_size_diff-expected_size_tree, num_changes);
+          octomap_msgs::fullMapToMsg(*octree, scene_msg.world.octomap.octomap);
+          if(scene_msg.world.octomap.octomap.id != OCTOMAP_MSG_TYPE) {
+            logWarn("fullMapToMsg produced unexpected octomap type: %s",
+                    scene_msg.world.octomap.octomap.id.c_str());
+          }
```
v4hn, line 871: see the comment below

planning_scene/src/planning_scene.cpp
```diff
@@ -1245,60 +1309,101 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
...
+  for (std::size_t i = 0; i < object_ids.size(); ++i)
+    if (object_ids[i] != OCTOMAP_NS)
+      world_->removeObject(object_ids[i]);
+}
 
-  if (map.data.empty())
-    return;
+void planning_scene::PlanningScene::processOctomapMsg(const octomap_msgs::OctomapWithPose &msg)
+{
+  const Eigen::Affine3d &t = getTransforms().getTransform(msg.header.frame_id);
```
v4hn, line 1322: This does not check if `frame_id.empty()` and could therefore print an error message.
Please check for that here and use `Identity` if the id is not set.
Until now it seems to be valid to send an empty octomap without a frame_id to remove it from the scene.

planning_scene/src/planning_scene.cpp
```diff
@@ -1245,60 +1309,101 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
...
-    const Eigen::Affine3d &t = getTransforms().getTransform(map.header.frame_id);
+    world_->removeObject(OCTOMAP_NS);
+  }
+  else if (msg.id == OCTOMAP_DIFF_MSG_TYPE)
+  {
+    collision_detection::CollisionWorld::ObjectConstPtr map = world_->getObject(OCTOMAP_NS);
+    if (map && map->shapes_.size() == 1)
+    {
+      if (msg.data.empty())
+        return;
```
v4hn, line 1354: Please move this check to the beginning of the surrounding block, so that the first thing that happens for empty diffs is check-and-return. There's no reason not to do that iif `map->shapes_.size() != 1` :)

planning_scene/src/planning_scene.cpp
```diff
@@ -1245,60 +1309,101 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
...
+      const shapes::OcTree *o = static_cast<const shapes::OcTree*>(map->shapes_[0].get());
+      boost::shared_ptr<const octomap::OcTree> octree = o->octree;
+      boost::shared_ptr<octomap::OcTree> nc_octree = boost::const_pointer_cast<octomap::OcTree>(octree);
```
v4hn, line 1357: :-1:
removing const-ness in a context that is this complex invites disaster.
If you wish to be able to modify the octree, keep a non-const pointer in the planning scene.
This would also allow for some further simplifications...
However, this would break ABI and therefore would only be possible for kinetic-devel (I see dave jumping happily up and down when reading this :P).
As an alternative, you could simply copy the old octomap and drop it. I would still prefer the other solution.

planning_scene/src/planning_scene.cpp
```diff
@@ -1245,60 +1309,101 @@ bool planning_scene::PlanningScene::usePlanningSceneMsg(const moveit_msgs::Plann
...
+  int expected_size = sizeof(int)+num_changes*((3*sizeof(unsigned short int))+sizeof(float));
+  if (expected_size > msg.data.size())
   {
-    logError("Received octomap is of type '%s' but type 'OcTree' is expected.", map.octomap.id.c_str());
+    logError("Did not receive enough data for specified diff size: %i bytes expected, %i received", expected_size, msg.data.size());
     return;
   }
+  if(expected_size < msg.data.size()) {
+      logWarn("Got more data than expected (%zu > %d)", msg.data.size(), expected_size);
+  }
```
v4hn, line 1392: Please simplify this into
`if(expected_size != msg.data.size()){ ...complain... }`.
MoveIt is way to verbose already.

v4hn:
> I wish we would all move away from deep changes like this in an LTS (indigo) release...

Long-Term-Support means Stable API / Stable ABI. It does _not_ mean "no new features".
As most people using MoveIt use indigo, it totally makes sense to add interesting features,
and also next year I don't see a reason not to consider merging new features in LTS branches.

planning_scene/src/planning_scene.cpp
davetcoleman, line 893: The ROS standard is 120 per style guidelines

davetcoleman:
I agree it does not mean no new features, but given our lack of good unit tests and integration tests its very risky we'll break current user's setups - for example it seems like this week's changes [are causing problems](https://github.com/ros-planning/moveit_ros/pull/716#issuecomment-234308320). I'm fine making mistakes, but it should be in the right branch.

planning_scene/src/planning_scene.cpp
davetcoleman, line 1392: @v4hn consider tweaking your rosconsole output to hide logging you are not interested in. i'm a big fan of good warnings like this so that when something goes wrong, we understand. there are too many instances and reports from users of moveit saying "failed to plan" with no other feedback. with a complex system like this we need as good tools to debug issues. see [my rosconsole article](http://dav.ee/blog/notes/archives/898) for tips

v4hn, line 1392: Thanks for the link! It's in my TODO queue.

I fully agree that MoveIt needs informative warnings and debug messages.
In cases like this one I would even prefer fatal errors that abort the system,
just to enforce the user actually setup a reasonable system that satisfies our assumptions
about the input.

On the other hand one always has to look at the amount of code bloat added because of log output
and this should really be minimal. It's hard enough already to read through existing functions
in MoveIt of 100+ lines where 90 of them are related to sanity checks and do not provide _any_
functionality.

In this particular case I believe it is perfectly fine to have _one_ warning
"Octomap diff contains unexpected amount of data: %i bytes expected, %i bytes received".
Do you disagree?

v4hn:
On Thu, Jul 21, 2016 at 01:18:03PM -0700, Dave Coleman wrote:

> I agree it does not mean no new features, but given our lack of good unit tests and integration tests its very risky we'll break current user's setups  for example it seems like this week's changes [are causing problems](https://github.com/ros-planning/moveit_ros/pull/716#issuecomment-234308320).
> I'm fine making mistakes, but it should be in the right branch.

I agree. But in my opinion neither `indigo-devel`, nor `kinetic-devel` are the right branch.
I would like to add this to next weeks agenda if we have time for it.

Btw: the message you referenced is not actually a problem of the indigo-devel branch.
Robert noticed an issue that was worse before, so this is still an improvement. :-)
I believe the biggest problem with the patch-set is https://github.com/ros-planning/moveit_ros/pull/716#discussion_r71799182 .

planning_scene/src/planning_scene.cpp
dornhege, line 1392: Actually that seems reasonable. If the data size is not what is expected (either less or more) something is probably wrong.

I would not go for anything that aborts without the option to only enable this globally. This could destroy a working system because of "something".

davetcoleman, line 1392: > It's hard enough already to read through existing functions in MoveIt of 100+ lines where 90 of them are related to sanity checks and do not provide _any_ functionality.

Yes its hard to read, but seeing that many sanity checks has not been my experience with the code base

davetcoleman:
The right branch is technically the ROS-L branch ;-)

v4hn:
On Fri, Jul 22, 2016 at 11:58:18AM -0700, Dave Coleman wrote:

> The right branch is technically the ROS-L branch ;-)

I would prefer a plain "master" branch.

planning_scene/src/planning_scene.cpp
dornhege, line 1357: This _should_ be safe, as it is happening in the same function that - without the diff function - would just replace the complete OctoMap.
I would like to avoid keeping a non-const pointer in the planning scene as that would have to be managed to. The only cleaner solution that I can see is to copy the old one, apply the diff on that and set that as the new octomap. This is less efficient, but will still be a win as the network overhead is avoided by the diff.

v4hn, line 1357: > This should be safe, as it is happening in the same function that - without the diff function - would just replace the complete OctoMap.

I don't see why this makes it safe to unconst a pointer out of a list of const elements from a class instance that has been extracted from a static_cast'ed const pointer from a map of objects that is from outside the world only const-visible.

Also I believe this does not involve much management to have a separate shared_ptr for the latest octree.
If the current message is a diff -> apply it to the maintained `octomap::OcTree` (and maybe still build a new `CollisionWorld::Object` with it?), otherwise generate a new octree, overwrite the current pointer and hand it down to the scene.

If you want to have this in indigo, then I would propose to have the copy in indigo and jade and add the pointer-version in kinetic.

dornhege, line 1357: I see what you mean. The clean way would be to remove the object from the scene, do the operations and re-insert the changed one. Currently the order is perform operations (with const-cast), remove, add. However, I don't see a way in the moveit API to extract something from a shape, while retaining it exclusively.

Having a second OcTree sounds hard as the actual OcTree lives in the CollisionWorld. There might be changes from different components in there (e.g. world updates and filtering by other objects) that would need to be managed.

The only somewhat simple and safe way that I see right now is to clone the OcTree and apply changes to that.

davetcoleman:
it'd be great to get this merged in before friday's merge!

planning_scene/src/planning_scene.cpp
v4hn, line 1357: On Mon, Aug 01, 2016 at 07:01:07AM -0700, Christian Dornhege wrote:

> I see what you mean. The clean way would be to remove the object from the scene,
> do the operations and re-insert the changed one.

Sadly, this is not possible without const-cast with the current API...

> There might be changes from different components in there (e.g. world updates and filtering by other objects) that would need to be managed.

Crap, I didn't think about that, you're right.

> The only somewhat simple and safe way that I see right now is to clone the OcTree and apply changes to that.

+1 please do that for now. If you could also address the other changes I'll +1 this for indigo.
It would be nice to have that in before Friday, I agree.

If you want to have an optimized version in MoveIt! (that would be really nice),
this will need some API revision in the World to access the non-const CollisionObject.

dornhege:
@130s : Do you want to address the proposed changes before the merge? Otherwise I can re-open a PR based on this one.

130s:
@dornhege thanks for the headsup and offering. Yes I appreciate if you could open a new PR.

dornhege:
I can currently not test as ros-planning/moveit_ros#736 is preventing this from working. This might be resolved quickly, if it has something to do with my setup or is easily fixable.
I could test with ros-planning/moveit_ros#ae51d07d5b1b848d87acb8c7dacb078b17d9c5ca and still provide this PR cleaned, if someone feels confident with merging. These should be independent as one is in moveit_core and the other in moveit_ros.

dornhege:
Continued in #322.

PR #306, opened 2016-07-07
Reduced excessive console output during tests

davetcoleman:
When running moveit_core tests there was a ton of output that slowed down the process - now it is optional through in-code verbose flag.

### DIFF
--- a/constraint_samplers/test/pr2_arm_kinematics_plugin.cpp
+++ b/constraint_samplers/test/pr2_arm_kinematics_plugin.cpp
@@ -105,11 +105,13 @@ int PR2ArmIKSolver::CartToJnt(const KDL::JntArray& q_init,
                               const KDL::Frame& p_in,
                               KDL::JntArray &q_out)
 {
+  const bool verbose = false;
   Eigen::Matrix4f b = KDLToEigenMatrix(p_in);
   std::vector<std::vector<double> > solution_ik;
   if(free_angle_ == 0)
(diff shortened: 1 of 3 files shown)


constraint_samplers/test/pr2_arm_kinematics_plugin.cpp
```diff
@@ -105,11 +105,13 @@ int PR2ArmIKSolver::CartToJnt(const KDL::JntArray& q_init,
                               const KDL::Frame& p_in,
                               KDL::JntArray &q_out)
 {
+  bool verbose = false;
```
v4hn: please make that `const bool verbose` everywhere.
This makes it clear it's no runtime option.

davetcoleman: good point, done

v4hn:
merged in f7e9d4ec267a82a030e48db2ebd78d73f7fa2d2d.
I suppose github does not pick up on requests that are merged as single commit from the command line.
Too bad, that would be a nice feature, if they already support "squash and merge" in their gui...

I suppose the same issue applies to indigo? +1 to cherry-pick it there.

davetcoleman:
Already merged through command line by @v4hn

v4hn:
Ah, yes. I left it open to discuss cherry-picks.
But I see you merged it into indigo and kinetic too.

Issue #310, opened 2016-07-14
C++11 Support for all moveit_* repos


davetcoleman:
It's [come to my attention](https://github.com/davetcoleman/moveit_visual_tools/issues/9) from @simonschmeisser that there is a linking error from any package that wants to use geometric_shapes with C++11 because of boost::variant c++11/c++03 incompatibilities. Because kinetic minimum requirements includes C++11, I think we should upgrade geometric_shapes to C++11. However, this would also require use to upgrade moveit_core since it depends on geometric_shapes. Which in turn would require us to upgrade all of MoveIt!, I believe. This is necessary sooner or later anyway. Thoughts? We've already upgraded moveit_planners here https://github.com/ros-planning/moveit_planners/pull/73 because of OMPL

simonschmeisser:
fcl also moved to c++11 in fcl > 0.4.0 and thus adds a c++11 requirement on moveit_core. fcl > 0.4.0 is required as fcl <= 0.4.0 is not compatible to octomap 1.8 as released in kinetic

https://github.com/flexible-collision-library/fcl/issues/137

v4hn:
+1 just add the respective compiler option to the repos.

davetcoleman:
The three above PRs for C++11 are ready to be merged together (geometric_shapes, moveit_core, moveit_ros)

rhaschke:
@davetcoleman Go ahead and merge those PRs whenever you see fit.

davetcoleman:
I have merged the three PRs

PR #311, opened 2016-07-15
Switched to C++11

davetcoleman:
Per https://github.com/ros-planning/moveit_core/issues/310

Note we should setup all the PRs across moveit_\* packages and merge all at once.


davetcoleman:
Travis will not pass until https://github.com/ros-planning/geometric_shapes/pull/44 is merged

PR #312, opened 2016-07-15
Add object types when requesting attached bodies in planning scene msg

dornhege:
Object types for attached bodies are currently not retrieved, when requesting a planning scene msg.     8db6548 fixes this.

In addition attached objects' types and colors are never pushed in pushDiffs, i.e., whenever pushDiffs was called, this information is lost for all attached bodies. 9ac5ca3 is added for that.

### DIFF
--- a/planning_scene/src/planning_scene.cpp
+++ b/planning_scene/src/planning_scene.cpp
@@ -471,7 +471,20 @@ void planning_scene::PlanningScene::pushDiffs(const PlanningScenePtr &scene)
     scene->getTransformsNonConst().setAllTransforms(ftf_->getAllTransforms());
 
   if (kstate_)
+  {
     scene->getCurrentStateNonConst() = *kstate_;
+    // push colors and types for attached objects
+    std::vector<const moveit::core::AttachedBody*> attached_objs;
+    kstate_->getAttachedBodies(attached_objs);
+    for(std::vector<const moveit::core::AttachedBody*>::const_iterator it = attached_objs.begin();
+            it != attached_objs.end(); ++it)
+    {
+        if(hasObjectType((*it)->getName()))
+            scene->setObjectType((*it)->getName(), getObjectType((*it)->getName()));
+        if(hasObjectColor((*it)->getName()))
+            scene->setObjectColor((*it)->getName(), getObjectColor((*it)->getName()));
+    }
+  }
 
   if (acm_)
     scene->getAllowedCollisionMatrixNonConst() = *acm_;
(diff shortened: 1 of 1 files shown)


planning_scene/src/planning_scene.cpp
```diff
@@ -909,11 +909,18 @@ void planning_scene::PlanningScene::getPlanningSceneMsg(moveit_msgs::PlanningSce
   if (comp.components & moveit_msgs::PlanningSceneComponents::TRANSFORMS)
     getTransforms().copyTransforms(scene_msg.fixed_frame_transforms);
 
-  if (comp.components & moveit_msgs::PlanningSceneComponents::ROBOT_STATE_ATTACHED_OBJECTS)
+  if (comp.components & moveit_msgs::PlanningSceneComponents::ROBOT_STATE_ATTACHED_OBJECTS) {
```
davetcoleman: brackets on new lines ;-)

davetcoleman:
+1 after style cleanup

dornhege:
Done.

planning_scene/src/planning_scene.cpp
```diff
@@ -910,10 +923,22 @@ void planning_scene::PlanningScene::getPlanningSceneMsg(moveit_msgs::PlanningSce
     getTransforms().copyTransforms(scene_msg.fixed_frame_transforms);
 
   if (comp.components & moveit_msgs::PlanningSceneComponents::ROBOT_STATE_ATTACHED_OBJECTS)
+  {
     robot_state::robotStateToRobotStateMsg(getCurrentState(), scene_msg.robot_state, true);
-  else
-    if (comp.components & moveit_msgs::PlanningSceneComponents::ROBOT_STATE)
+    for(std::vector<moveit_msgs::AttachedCollisionObject>::iterator it
+            = scene_msg.robot_state.attached_collision_objects.begin();
```
v4hn: Please remove that line-break. That's the most weird way of breaking in a for-statement I've seen so far..

dornhege: Also done.

v4hn:
ok full +1, cherry-pick to j/k.
Now, I _would_ merge, but nobody else reviewed your second (totally unrelated) commit to change `pushDiff`. So I'll leave the merge to the next one.
Please open a new request for orthogonal changes next time.
Thanks for the request!

dornhege:
The second commit is essential for this to work. It is the same problem: object types and colors were only handled for CollisionObjects, but not attached ones. Without the pushDiff fix, these will always be cleared and the retrieval in getPlanningSceneMsg fails.

davetcoleman:
I agree with @v4hn you shouldn't add a different change to an already reviewed PR without at least announcing you have done so in the comments

dornhege:
Both commits were there from the start. I only amended the style changes to the latter one, which might have confused you.

v4hn:
ok, so its Github's fault then. Sorry for blaming you @dornhege :-)
Thanks again!

davetcoleman:
Sorry for the confusion @dornhege 

I've cherry-picked to jade and kinetic

PR #314, opened 2016-07-25
Add a missing #include <boost/bind.hpp>

de-vri-es:
This PR adds a missing `#include <boost/bind.hpp>` which caused a compilation failure on Arch Linux.


v4hn:
This is an obvious bug.
Thanks for reporting @de-vri-es !

v4hn:
cherry-picked to indigo-devel and jade-devel

PR #317, opened 2016-07-26
catkin_lint cleanup jade-devel

v4hn:
This is basically #305 for jade.


davetcoleman:
why is travis not running?

v4hn:
No idea, it works for #318 . Any idea how to trigger it?

d130s:
It can happen between Travis and Github. Close the pr and reopen it might
work (I can't for now from my phone).

davetcoleman:
It doesn't appear on Travis... maybe it is to similar to #318 so it filtered it out? Attempting to reopen.

davetcoleman:
Travis: Could not find a package configuration file provided by "Eigen3"

v4hn:
okok, I yield :p .
I thought `find_package(Eigen3)` should be fine in jade,
but this still hits this weird problem that catkin tools can't find Eigen3 although `FindEigen3.cmake` is around in ubuntu 14.04..
Anyway, I updated the request.

PR #322, opened 2016-08-03
Pull302 fixup

dornhege:
Continued from #302.

Addresses review comments.


mikeferguson:
@dornhege -- unfortunately it looks like this did not get merged before the moveit repos got consolidated -- if you reopen against https://github.com/ros-planning/moveit we can review and merge there

PR #323, opened 2016-08-05
Added notice repo has been migrated to github.com/ros-planning/moveit

davetcoleman:
Please review and if appropriate I'd like a +1 that we can apply this change to the README.md for all deprecated repos tomorrow.


v4hn:
+1 This also has to be picked to indigo and jade

davetcoleman:
Done for all branches and repos

